PR #9, opened 2026-01-07
[slack] add support for typing, status indicators in slack adapter

visyat:
Adds support for typing indicator in Slack


nandorojo:
@visayat why'd we close this btw?

PR #29, opened 2026-02-05
refactor: replace broad googleapis dependency with specific packages

RihanArfan:
Switches from `googleapis` to more specific dependencies to reduce the output bundle size when using the gchat adapter


haydenbleasel:
@RihanArfan updated ye branch 👍

PR #34, opened 2026-02-09
Add multi-workspace support to slack adapter

NicolasMontone:
- removes the hardcoded botToken
- uses async local storage to get token per teamId
- added `slackAdapter.withBotToken` to run it outside request 
- allow saving installations encrypted in the kv
- added handleOAuthCallback to handle installations

test; https://vercel.slack.com/archives/C08077A6JDB/p1770674710628159?thread_ts=1770671171.096509&cid=C08077A6JDB

### DIFF
--- a/.changeset/rare-houses-greet.md
+++ b/.changeset/rare-houses-greet.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": minor
+---
+
+Add multi-workspace support. A single Slack adapter instance can now serve multiple workspaces by resolving bot tokens per-request via AsyncLocalStorage. Includes OAuth V2 flow handling, installation management (set/get/delete), optional AES-256-GCM token encryption at rest, and a withBotToken helper for out-of-webhook contexts


packages/adapter-slack/src/index.ts
```diff
@@ -239,6 +306,205 @@ export class SlackAdapter implements Adapter<SlackThreadId, unknown> {
...
+      this.logger.info("Slack adapter initialized in multi-workspace mode");
+    }
+  }
+
+  // ===========================================================================
+  // Multi-workspace installation management
+  // ===========================================================================
+
+  private installationKey(teamId: string): string {
+    return `slack:installation:${teamId}`;
```
malewis5, line 320: @NicolasMontone does this support both org level installs and workspace level installs.

malewis5: You also can save [user tokens](https://docs.slack.dev/authentication/tokens/#user) for an installation.

slack:installation:${teamId}:${userId}

User tokens are helpful when you need access to like a DM or something that bots aren't allowed to access.

NicolasMontone: I think this is workspace?

Issue #113, opened 2026-02-25
ESM directory import fails at runtime — ERR_UNSUPPORTED_DIR_IMPORT


hkbertoson:
### Bug Description

`@chat-adapter/teams` is "type": "module" and the compiled output in dist/index.js contains a bare directory import:

```
from "@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials"
```

Node.js ESM does not support directory imports. This fails at runtime with:
```
ERR_UNSUPPORTED_DIR_IMPORT: Directory import '…/authProviders/azureTokenCredentials' is not supported
```

Update the import to include the full file path: "@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials/index.js"
Or configure tsup to bundle this dependency so the bare import doesn't appear in the output

### Steps to Reproduce

Create a new chatbot with the teams adapter

### Expected Behavior

It loads the bot

### Actual Behavior

Not loading the bot

### Code Sample

```typescript

```

### Chat SDK Version

Latest

### Platform Adapter

Microsoft Teams

### Operating System

macOS

PR #121, opened 2026-02-26
fix(teams): avoid ESM directory import for graph auth provider

cipher416:
Fixes vercel/chat#113.

Node.js ESM does not support bare directory imports. The Teams adapter compiled output included:
\n- "@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials"\n
which throws ERR_UNSUPPORTED_DIR_IMPORT at runtime.

Change the import to the explicit file path:
\n- "@microsoft/microsoft-graph-client/authProviders/azureTokenCredentials/index.js"\n
Validation:
- Reproduced failure by installing @chat-adapter/teams in a fresh npm project and importing it via Node ESM.
- Built @chat-adapter/teams and verified dist output contains the fixed path.
- Packed and installed the tarball; import succeeds.

Also ran: pnpm check / pnpm fix (no changes).

### DIFF
--- a/packages/adapter-teams/src/index.test.ts
+++ b/packages/adapter-teams/src/index.test.ts
@@ -1,3 +1,6 @@
+import { execSync } from "node:child_process";
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
 import type { Logger } from "chat";
 import { describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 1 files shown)


haydenbleasel:
nice one @cipher416 - i also added a unit test to ensure future changes don't break node ESM imports.

PR #124, opened 2026-02-26
feat(mcp): Introduce MCP connection to chat

sergical:
## Summary

- **Per-call auth headers**: `callTool()` now accepts an optional `options.headers` parameter for per-user/per-tenant authentication. When provided, a temporary MCP client+transport is created with the override headers for the duration of the call.
- **MCP feature docs**: New `mcp.mdx` guide covering configuration, tool listing/calling, transport types, three auth patterns (service-level, per-tenant, per-user), multiple servers, and error codes.
- **API reference updates**: Added `mcpServers` to the `Chat` constructor table and a `mcp` property section.

## Test plan

- [x] 19/19 MCP tests pass (`pnpm --filter chat test src/mcp.test.ts`)
- [x] 466/466 total tests pass
- [x] TypeScript type-check passes
- [x] Lint/format passes
- [ ] Preview docs locally (`pnpm --filter docs dev` → `/docs/mcp`)

## Prompts
* https://www.traces.com/s/jn7963brr6pptn6mfna4fxsy0d81xf7f
* https://www.traces.com/s/jn7ez6k9atszs87admsv74bhhd81xck3
* https://www.traces.com/s/jn74v0affax2xdvkb87zpc24q181x2kf

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/docs/content/docs/mcp.mdx
+++ b/apps/docs/content/docs/mcp.mdx
@@ -79,7 +79,7 @@ bot.onNewMention(async (thread, message) => {
 
 ## Using with AI SDK
 
-If you're building an AI-powered bot with [AI SDK](https://ai-sdk.dev), use [`@ai-sdk/mcp`](https://ai-sdk.dev/docs/ai-sdk-core/mcp-clients) to let the model decide which MCP tools to call. This is complementary to the built-in `bot.mcp` client — use `@ai-sdk/mcp` when an LLM should choose tools, and `bot.mcp` when your code should call tools directly.
+If you're building an AI-powered bot with [AI SDK](https://ai-sdk.dev), use [`@ai-sdk/mcp`](https://ai-sdk.dev/docs/ai-sdk-core/mcp-tools) to let the model decide which MCP tools to call. This is complementary to the built-in `bot.mcp` client — use `@ai-sdk/mcp` when an LLM should choose tools, and `bot.mcp` when your code should call tools directly.
 
 ```bash
 pnpm add @ai-sdk/mcp ai
@@ -87,7 +87,7 @@ pnpm add @ai-sdk/mcp ai
 
 Create an MCP client and pass its tools to a `ToolLoopAgent`:
 
-<CodeBlockTabs>
+<CodeBlockTabs defaultValue="gateway">
   <CodeBlockTabsList>
     <CodeBlockTabsTrigger value="gateway">AI Gateway</CodeBlockTabsTrigger>
     <CodeBlockTabsTrigger value="anthropic">Anthropic</CodeBlockTabsTrigger>


haydenbleasel:
This is amazing. Thinking more about how we could tie MCP into Chat SDK overall — there seem to be two paths here and I'd love your extended thoughts on the tradeoffs.

### Option 1: MCP-native (this PR)

Chat SDK owns the MCP client lifecycle, connection management, multi-server aggregation, per-call auth, etc. Users interact with MCP through `bot.mcp.listTools()` / `bot.mcp.callTool()`.

### Option 2: MCP-via-AI-SDK

AI SDK already has MCP support via `@ai-sdk/mcp`. Users would manage MCP clients themselves and pass discovered tools into `generateText`/`streamText`:

```ts
import { createMCPClient } from '@ai-sdk/mcp';
import { generateText } from 'ai';

const mcpClient = await createMCPClient({
  transport: { type: 'sse', url: 'https://jira-mcp.example.com/mcp' },
});

  const result = await generateText({
    model: anthropic('claude-sonnet-4-5'),
    tools: await mcpClient.tools(),
    prompt: message.text,
  });
  await thread.post(result.text);
});
```

### Thinking out loud

The MCP-native approach has clear benefits around lifecycle management, graceful failure, and per-call auth headers — but I'm curious what deeper integrations you envision that would make Chat SDK the right layer to own this vs. letting AI SDK handle it. For example, could Chat SDK auto-expose MCP tools as slash commands, or route tool calls differently per platform?

Would love to hear your thoughts on where this could go — it could be very valuable to have MCP deeply integrated at the Chat SDK level if there are platform-aware features that AI SDK alone can't provide.

sergical:
Great questions @haydenbleasel! You're right that there are two paths — and they're complementary, not competing.

### When to use which

**`bot.mcp` (MCP-native, this PR):**
- ✅ Your code decides which tools to call (deterministic workflows)
- ✅ Chat SDK manages connection lifecycle, graceful degradation, multi-server aggregation
- ✅ Per-call auth headers (tenant/user-scoped tokens)
- ✅ No AI/LLM dependency — works for slash commands, keyword triggers, etc.
- Good for: "when user says `/sentry <query>`, call `search_issues`"

**`@ai-sdk/mcp` (MCP-via-AI-SDK):**
- ✅ The LLM decides which tools to call based on the message
- ✅ Works with any AI SDK provider (Anthropic, OpenAI, AI Gateway, etc.)
- ✅ Natural multi-tool orchestration — model chains tools as needed
- ⚠️ Users manage MCP client lifecycle themselves
- Good for: "let the model figure out if it should search issues, create a ticket, or look up a user"

### They compose well together

A bot can use both in the same app — `bot.mcp` for deterministic handlers and `@ai-sdk/mcp` for AI-powered ones. Chat SDK doesn't need to own the AI SDK MCP client because `thread.post(result.textStream)` already handles streaming from any AI SDK response.

### Deeper integrations (future)

Your instinct about platform-aware features is spot on. Some things that could be interesting down the line:
- Auto-exposing MCP tools as slash commands (register tool names as `/tool-name`)
- Platform-specific tool rendering (e.g. Sentry issues as rich Slack cards vs plain text in Teams)
- Tool result caching per-thread via the state adapter

For now I've added a ["Using with AI SDK" section](https://github.com/vercel/chat/blob/feat/mcp-client-support/apps/docs/content/docs/mcp.mdx) to the MCP docs showing the `@ai-sdk/mcp` pattern with tabbed provider examples, plus a callout comparing when to use each approach.

bensabic:
Thank you for the work on this PR @sergical! This was discussed internally and ultimately adding MCP support to the Chat SDK isn't a direction we'll be taking, for now.

PR #128, opened 2026-02-27
fix(discord): add Partials.Channel to gateway client for DM support

iamtraction:
The Discord adapter's `runGatewayListener` creates a discord.js `Client` with `GatewayIntentBits.DirectMessages` but without `Partials.Channel`. discord.js requires this partial to receive DM events. Without it, DM `MessageCreate` events are silently dropped.

### DIFF
--- a/packages/adapter-discord/src/gateway.test.ts
+++ b/packages/adapter-discord/src/gateway.test.ts
@@ -16,7 +16,9 @@ const { MockClient, mockClientInstance } = vi.hoisted(() => {
     destroy: vi.fn(),
     user: { username: "testbot", id: "bot123" },
   };
-  const MockClient = vi.fn().mockImplementation(function (this: typeof mockClientInstance) {
+  const MockClient = vi.fn().mockImplementation(function (
+    this: typeof mockClientInstance
+  ) {
     Object.assign(this, mockClientInstance);
     return this;
   });
@@ -76,9 +78,13 @@ describe("Gateway client configuration", () => {
 
     let listenerPromise: Promise<unknown> | undefined;
     const response = await adapter.startGatewayListener(
-      { waitUntil: (p) => { listenerPromise = p as Promise<unknown>; } },
+      {
+        waitUntil: (p) => {
+          listenerPromise = p as Promise<unknown>;
+        },
+      },
       1000,
-      controller.signal,
+      controller.signal
     );
 
     expect(response.status).toBe(200);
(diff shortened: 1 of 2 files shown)


haydenbleasel:
Investigated this and can confirm the fix is correct. Here's the evidence:

## The mechanism (discord.js v14 source)

`MessageCreate.js` does this when a message arrives:

```js
const channel = this.getChannel({ id: data.channel_id, ... });
if (channel) {
  // ... emit Events.MessageCreate
}
```

`Action.js`'s `getPayload` (which `getChannel` calls) does:

```js
getPayload(data, manager, id, partialType, cache) {
  return this.client.options.partials.includes(partialType)
    ? manager._add(data, cache)    // Partials.Channel enabled → construct channel
    : manager.cache.get(id);       // Not enabled → lookup cache (returns undefined for DMs)
}
```

Discord's API stopped sending `CHANNEL_CREATE` for DM channels, so they're never in the cache. Without `Partials.Channel`, `getPayload` returns `undefined`, the `if (channel)` check fails, and the event is silently dropped — no error, no warning.

## Maintainer confirmation

- **[discord.js#5516](https://github.com/discordjs/discord.js/issues/5516)** — `monbrey` (discord.js team): *"The v8 Discord API/gateway does not emit CHANNEL_CREATE for DM Channels. As a result, they don't exist in discord.js cache and messages won't emit for them. Enabling the CHANNEL partial should fix the issue."*

- **[discord.js#7699](https://github.com/discordjs/discord.js/issues/7699)** — `monbrey`: *"DM channels don't fire the channelCreate event which means they don't get placed in the discord.js cache. As the channel is uncached, messageCreate events for it won't fire unless you enable the Channel partial also."*

- **[discord.js#7487](https://github.com/discordjs/discord.js/issues/7487)** — `kyranet` (discord.js team) in the design roadmap: *"Discord stopped sending us DM channels during the READY dispatch... if somebody wanted to receive messages from DMs, not only they would need to specify the intent, but they would need to find out that a second step is needed: specifying Partials.Channels."*

## TL;DR

Both `GatewayIntentBits.DirectMessages` **and** `Partials.Channel` are required. The current code has the intent but not the partial, so DM events are silently dropped at runtime. Types and builds pass fine because it's a runtime cache lookup issue, not a type error. Fix is correct and minimal.

haydenbleasel:
Thanks @iamtraction great catch 🙏

PR #139, opened 2026-02-28
feat(chat): allow disabling fallback streaming placeholder

vincenzodomina:
## Summary
- Add `fallbackStreamingPlaceholderText?: string | null` to `ChatConfig`.
- Preserve `null` through `Chat` and `ThreadImpl` so fallback streaming can skip the initial `"..."` post.
- Keep the default behavior unchanged: omitting the option still posts `"..."` immediately.

## Motivation
Issue #140 only needs a way to suppress the initial placeholder on post+edit fallback adapters (for example Telegram), so this PR keeps the change scoped to that behavior.

When the placeholder is `null`, the first real streamed chunk creates the message and the existing edit loop continues from there.

See issue: https://github.com/vercel/chat/issues/140

## Test plan
- `pnpm --dir packages/chat exec vitest run src/thread.test.ts src/chat.test.ts`
- `pnpm --dir packages/chat run typecheck`

### DIFF
--- a/apps/docs/content/docs/streaming.mdx
+++ b/apps/docs/content/docs/streaming.mdx
@@ -58,6 +58,26 @@ const bot = new Chat({
 });
 ```
 
+### Disabling the placeholder message
+
+By default, post+edit adapters send an initial `"..."` placeholder message before the first chunk arrives. You can disable this to wait for real content before posting:
+
+```typescript title="lib/bot.ts" lineNumbers
+const bot = new Chat({
+  // ...
+  fallbackStreamingPlaceholderText: null,
+});
+```
+
+You can also customize the placeholder text:
+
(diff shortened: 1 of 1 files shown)


haydenbleasel:
Thanks for the contribution @vincenzodomina! 🙌

I pushed a few small fixes on top of your work:

- Added a missing test for the empty stream + `placeholder=null` edge case
- Replaced the `(customChat as any).createThread` test with one that exercises the full Chat → Thread pipeline through public APIs (less fragile)
- Sorted interface members alphabetically to pass the Ultracite linter
- Added a changeset
- Updated the docs

Full validation passes — should be good to go!

vincenzodomina:
Whoa thanks for the fast merge!

PR #142, opened 2026-02-28
fix: use createMemoryState if no REDIS_URL

nicopujia:
## Summary

Update `examples/nextjs-chat` to use in-memory state adapter if `REDIS_URL` isn't set. 

Note that the env var is optional, as the [docs](https://github.com/vercel/chat/tree/main/examples/nextjs-chat#environment-variables) imply:

> At minimum, set `BOT_USERNAME` and credentials for one platform

So it shouldn't be mandatory.

## Test plan

- [x] `pnpm validate`
- [x] Manual testing (ran the example with Discord variables set up and navigated to `/api/webhooks/discord`)


haydenbleasel:
Good idea, thanks @nicopujia 🚀

Issue #131, opened 2026-02-27
Discord: onSlashCommand handlers never invoked via HTTP Interactions


luminus-xd:
### Bug Description

`onSlashCommand` handlers registered via `bot.onSlashCommand(...)` are never invoked
when slash commands arrive through Discord's HTTP Interactions (webhook).

The `handleWebhook` method correctly returns `DeferredChannelMessageWithSource` to
acknowledge the interaction, but never calls `this.chat.processSlashCommand()`.
As a result, Discord shows "The application did not respond" after the timeout.

### Steps to Reproduce

1. Register a slash command handler:
   ```typescript
   bot.onSlashCommand("help", async (event) => {
     await event.channel.post("Hello!");
   });
   ```
2. Set Interactions Endpoint URL in the Discord Developer Portal
3. Invoke `/help` in Discord
4. Discord shows "The application did not respond"

### Root Cause

In `handleWebhook`, `MessageComponent` (button clicks) correctly calls
`handleComponentInteraction()` which invokes `this.chat.processAction()`.
However, the `ApplicationCommand` branch only returns a deferred response
without processing:

```javascript
// ✅ MessageComponent — handler is invoked
if (interaction.type === InteractionType.MessageComponent) {
  this.handleComponentInteraction(interaction, options); // calls processAction()
  return this.respondToInteraction({ type: InteractionResponseType.DeferredUpdateMessage });
}

// ❌ ApplicationCommand — handler is never invoked
if (interaction.type === InteractionType.ApplicationCommand) {
  return this.respondToInteraction({
    type: InteractionResponseType.DeferredChannelMessageWithSource,
  });
  // Missing: this.handleSlashCommandInteraction(interaction, options)
}
```

There is no `handleSlashCommandInteraction` method in `DiscordAdapter`,
and `this.chat.processSlashCommand()` is never called anywhere in the adapter.

### Expected Behavior

`onSlashCommand` handlers are invoked and their response is sent back to Discord
via the interaction follow-up API.

### Actual Behavior

The deferred response is sent but the handler is never called, resulting in
"The application did not respond" in the Discord client.

### Chat SDK Version

4.15.0

### Platform Adapter

Discord

PR #152, opened 2026-03-02
Ensure handling of Discord slash commands

PClmnt:
This PR updates the Discord Adapter to allow correctly handling of slash commands in Discord. Previously, slash commands would just be ACK'd and no response would take place. Also, Discord slash commands require defer + editing the original interaction response.

This updates the adapter so that: 

- APPLICATION_COMMAND events are now dispatched to processSlashCommand.
- First slash response updates @original (PATCH /webhooks/{app_id}/{token}/messages/@original).
- Subsequent responses use followups (POST /webhooks/{app_id}/{token}).
- Works for both text and file attachments in slash responses.

There's a question open, and i've left a comment in the code about how we might want to handle nested subcommands:

## Test plan 
- [ ] Bot receives a Discord slash command, adapter returns deferred ACK (type 5) immediately — should succeed
- [ ] Bot posts first response via event.channel.post(...) after defer — should replace deferred “thinking…” message (PATCH @original)
- [ ] Bot posts a second response in same slash-command flow — should be sent as a follow-up message (POST webhook)
- [ ] Bot posts a slash-command response with file attachments — should succeed via interaction webhook (no channel API fallback)
- [ ] Slash command invoked inside a Discord thread — should resolve deferred response correctly in that thread context
- [ ] Regular non-slash message posting/editing/deleting/reactions continue to work as before

## Video

https://github.com/user-attachments/assets/0f1b79cb-f50c-49d6-83c7-f4235b4f457b

## Screenshots
image
image
image

---

Resolves https://github.com/vercel/chat/issues/131

### DIFF
--- a/apps/docs/content/docs/adapters/discord.mdx
+++ b/apps/docs/content/docs/adapters/discord.mdx
@@ -174,6 +174,7 @@ CRON_SECRET=your-random-secret                   # For Gateway cron
 
 | Feature | Supported |
 |---------|-----------|
+| Slash Commands | Yes |
 | Mentions | Yes |
 | Reactions (add/remove) | Yes |
 | Cards (Embeds) | Yes |
--- a/apps/docs/content/docs/adapters/index.mdx
+++ b/apps/docs/content/docs/adapters/index.mdx
@@ -36,6 +36,7 @@ Adapters handle webhook verification, message parsing, and API calls for each pl
 
 | Feature | Slack | Teams | Google Chat | Discord | Telegram | GitHub | Linear |
 |---------|-------|-------|-------------|---------|----------|--------|--------|
+| Slash commands | ✅ | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ |
 | Mentions | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | Add reactions | ✅ | ❌ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | Remove reactions | ✅ | ❌ | ✅ | ✅ | ✅ | ⚠️ | ⚠️ |
--- a/apps/docs/content/docs/slash-commands.mdx
+++ b/apps/docs/content/docs/slash-commands.mdx
@@ -7,9 +7,12 @@ prerequisites:
 related:
   - /docs/modals
   - /docs/adapters/slack
+  - /docs/adapters/discord
 ---
 
 Slash commands let users invoke your bot with `/command` syntax. Register handlers with `onSlashCommand` to respond.
+
+Slash commands are supported on [Slack](/docs/adapters/slack) and [Discord](/docs/adapters/discord).
 
 ## Handle a specific command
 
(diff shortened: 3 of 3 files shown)


packages/adapter-discord/src/index.ts
```diff
@@ -366,6 +371,123 @@ export class DiscordAdapter implements Adapter<DiscordThreadId, unknown> {
...
+            channelId,
+          },
+          options
+        );
+      }
+    );
+  }
+
+  private formatSlashCommandText(options?: DiscordCommandOption[]): string {
+    // TODO(discord-slash): Subcommand and subcommand-group names are currently
```
PClmnt: Still an open question of what exactly to do here, not exactly sure what would be preferred

**Example payload:**

Command: /project issue create
Options:
title = "Login fails"
priority = "high"

**Current event:**
event.command = "/project"
event.text = "Login fails high"

**What we're missing:**
We're not actually mapping all the of the subcommands 


Maybe something like this? 
```
event.command = "/project issue create"
event.text = "title=\"Login fails\" priority=high"

```

haydenbleasel:
Hey @PClmnt! Great work on this PR 🙌 I pushed a few commits on top of yours:

- **Expand subcommand path into `event.command`** — `parseSlashCommand` now appends subcommand/group names to the command path (e.g. `/project issue create`), keeping `event.text` as flat values. This addresses the open question you left in the TODO comment.
- **Fix `initialResponseSent` race condition** — moved the flag set before `await` so concurrent `post()` calls don't both try to PATCH `@original`.
- **Restore unrelated indentation change** — reverted the whitespace-only diff in `handleComponentInteraction` to keep the PR focused.
- **Extract `tryPostSlashResponse` helper** — deduplicates the slash-context check that was repeated in `postMessage` and `postChannelMessage`.
- **Fix integration test mock** — added interaction webhook route handlers (`editOriginal`/`followUp`) to `setupDiscordFetchMock` so the slash command integration test actually works end-to-end.

Also updated docs.

PClmnt:
> Hey @PClmnt! Great work on this PR 🙌 I pushed a few commits on top of yours:
> 
> * **Expand subcommand path into `event.command`** — `parseSlashCommand` now appends subcommand/group names to the command path (e.g. `/project issue create`), keeping `event.text` as flat values. This addresses the open question you left in the TODO comment.
> * **Fix `initialResponseSent` race condition** — moved the flag set before `await` so concurrent `post()` calls don't both try to PATCH `@original`.
> * **Restore unrelated indentation change** — reverted the whitespace-only diff in `handleComponentInteraction` to keep the PR focused.
> * **Extract `tryPostSlashResponse` helper** — deduplicates the slash-context check that was repeated in `postMessage` and `postChannelMessage`.
> * **Fix integration test mock** — added interaction webhook route handlers (`editOriginal`/`followUp`) to `setupDiscordFetchMock` so the slash command integration test actually works end-to-end.
> 
> Also updated docs.

Thanks so much for taking the time to sort those issues out and getting this merged though, I really appreciate it! All looks good on my side. 🚢🚢

PR #154, opened 2026-03-02
Implement state adapter based on Postgres

bai:
## Summary

Postgres might not be an ideal state store for the use case, but for smaller apps it could be good enough to avoid adding redis to the stack.

## Test plan

Work in progress, going to deploy this to my chat app.


haydenbleasel:
This is great work — we'd love to have this as part of the core Chat SDK\! I'm going to jump in and help finish this up. Thanks for the contribution\! 🙌

haydenbleasel:
Migrated from `postgres` to `pg` (node-postgres) per @cramforce's request. All tests passing with 100% coverage, full validation green.

bai:
I got distracted and been meaning to finish on a weekend. Thanks for picking it up ❤️

PR #159, opened 2026-03-02
feat(telegram): support sendMessageDraft streaming in DMs

timolins:
## Summary
- add native `stream()` support in Telegram adapter using `sendMessageDraft` for private chats
- keep post+edit fallback for group/topic chats and for environments where `sendMessageDraft` is unavailable
- clean up placeholder messages for empty fallback streams and tighten unsupported-method detection
- add/expand Telegram streaming tests (DM draft path, non-DM fallback, unsupported method fallback)
- update docs/feature matrix to reflect `Draft API (DMs) + Post+Edit` behavior

## Verification
- `pnpm --filter @chat-adapter/telegram typecheck`
- `pnpm --filter @chat-adapter/telegram test`

## Notes
- Telegram `sendMessageDraft` was introduced in Bot API 9.5 and is currently DM-scoped in this adapter path.

### DIFF
--- a/packages/adapter-telegram/src/index.ts
+++ b/packages/adapter-telegram/src/index.ts
@@ -806,7 +806,10 @@ export class TelegramAdapter
 
       renderedAccumulated = this.renderStreamMarkdown(rawAccumulated);
 
-      if (!renderedAccumulated.trim() || renderedAccumulated === lastDraftText) {
+      if (
+        !renderedAccumulated.trim() ||
+        renderedAccumulated === lastDraftText
+      ) {
         continue;
       }
 


haydenbleasel:
Hey Timo! Great PR — the draft streaming approach is really clean. I pushed a few follow-up fixes after reviewing:

### What changed

1. **Added a comment explaining the `isDM` heuristic** — Telegram always assigns negative IDs to groups/supergroups/channels and positive IDs to private chats, so `!chatId.startsWith("-")` is reliable. Added a comment to make that explicit for future readers.

2. **Mid-stream draft failures now fall back to post+edit** — Previously, if `sendMessageDraft` failed *after* the first successful draft update, the adapter silently disabled draft streaming and the user saw nothing until the final `postMessage`. Now any draft failure (not just the first unsupported-method error) immediately falls back to `streamViaPostEdit` with the remaining chunks, so the streaming UX stays consistent.

3. **Deferred markdown rendering in the draft stream path** — `renderStreamMarkdown` was being called on every incoming chunk, even when the throttle interval hadn't elapsed. Restructured to check the time interval first and only render when we're actually going to send a draft update.

4. **Same rendering optimization in `streamViaPostEdit`** — Applied the same time-check-first pattern to the post+edit fallback path, skipping `renderStreamMarkdown` when the interval hasn't elapsed yet.

5. **Simplified `createDraftId`** — Replaced the verbose `typeof crypto !== "undefined" && typeof crypto.randomUUID === "function"` guard with `globalThis.crypto?.randomUUID?.()` and a nullish coalescing fallback.

All changes pass `pnpm validate` cleanly (typecheck, lint, tests, build). 👍

Humorist7475:
Hey, any update on the status of this PR? Would be nice to get this merged in.

bensabic:
Hey @timolins! Are you able to address the conflicts when you have a chance? Once you do, we'd be happy to review it once more and can have it merged.

bensabic:
Closing PR due to inactivity.

PR #185, opened 2026-03-05
fix discord thread starter root message parsing

dardanbujupaj:
## Summary
- Parse Discord `ThreadStarterMessage` (type `21`) using `referenced_message` when available, so root thread messages preserve original parent content and metadata.
- Keep graceful fallback behavior when `referenced_message` is missing or null.

Discord docs note that thread starter messages can omit content/attachments and primarily carry `message_reference` + optional `referenced_message`:
- https://discord.com/developers/docs/resources/message#message-reference-content-attribution-thread-starter-messages
- https://discord.com/developers/docs/resources/message#message-object-message-types

### Notes
Discord developer docs state that `...[If] referenced_message field is not present, the backend did not attempt to fetch the message that was being replied to, so its state is unknown.`. Not exactly sure from the wording, if that can happen with threads as well or only with normal replys. 
We could introduce a message fetch as fallback if `referenced_message` is undefined, but I think falling back to the thread start message is fine, unless we find out that undefined is actually possible in thread start messages. 
For deleted original messages (`referenced_message == null`), falling back to the thread start message is correct IMO.

I still use the original thread starting message as `raw`, not sure if that should better be the actual parsed message.

Open for opinions on these decisions :)

## Test plan
- `pnpm validate` passes
- Add regression tests for parser behavior and `listThreads` root-message handling.

### DIFF
--- a/packages/adapter-discord/src/index.ts
+++ b/packages/adapter-discord/src/index.ts
@@ -1480,7 +1480,7 @@ export class DiscordAdapter implements Adapter<DiscordThreadId, unknown> {
       threadId,
       text: this.formatConverter.extractPlainText(msg.content),
       formatted: this.formatConverter.toAst(msg.content),
-      raw: msg,
+      raw,
       author: {
         userId: author.id,
         userName: author.username,


haydenbleasel:
nice one @dardanbujupaj - appreciate it!

PR #189, opened 2026-03-05
fix: atomic message deduplication to prevent silent app_mention drops

### LINKED ISSUE #180: Slack: non-atomic message deduplication causes app_mention events to be silently dropped
### Bug Description

When a Slack app subscribes to both `app_mention` and `message.channels` events (as recommended in the SDK docs), Slack sends two events for every @mention with the same `ts`. The SDK's deduplication logic uses a non-atomic GET-then-SET pattern, allowing both events to pass the dedupe check simultaneously. When `message.channels` wins the race and gets processed first, it doesn't match `onNewMention` (since it's not typed as an `app_mention`), so no handler fires. The subsequent `app_mention` event is then either deduped or blocked by the thread lock — resulting in the bot silently ignoring the mention.

This causes intermittent failures where the bot responds to ~1 in 4 @mentions and silently drops the rest.

### Steps to Reproduce

1. Create a Slack app with the recommended event subscriptions: `app_mention`, `message.channels`, `message.groups`, `message.im`, `message.mpim`
2. Create a Chat instance with Redis state and Slack adapter (multi-workspace OAuth mode)
3. Register `onNewMention` and `onSubscribedMessage` handlers per the SDK docs
4. @mention the bot in a Slack channel
5. Observe that the bot intermittently does not respond (~75% failure rate on warm instances)

### Expected Behavior

Every @mention should trigger the `onNewMention` handler (or `onSubscribedMessage` if the thread is subscribed). The SDK should correctly deduplicate the two Slack events and ensure the `app_mention` event is the one that gets processed, since it carries the mention semantic.

### Actual Behavior

When `message.channels` wins the race to the dedupe key, the message is processed but doesn't match `onNewMention` (it's a regular message, not an app_mention). The `app_mention` event then arrives and is either:
- Silently deduped (key already set), or
- Blocked by the thread lock (LockError)

In both cases, the bot produces no response. Vercel logs show `POST 200` with zero handler log output for the failed menti

### Code Sample

```typescript
import { Chat } from "chat";
import { createSlackAdapter } from "@chat-adapter/slack";
import { createRedisState } from "@chat-adapter/state-redis";

const bot = new Chat({
  adapters: {
    slack: createSlackAdapter({
      clientId: process.env.SLACK_CLIENT_ID!,
      clientSecret: process.env.SLACK_CLIENT_SECRET!,
      signingSecret: process.env.SLACK_SIGNING_SECRET!,
    }),
  },
  state: createRedisState({ url: process.env.REDIS_URL! }),
});

  console.log("onNewMention fired:", message.text); // Often never logs
  await thread.subscribe();
  await thread.post("Hello!");
});

bot.onSubscribedMessage(async (thread, message) => {
  console.log("onSubscribedMessage fired:", message.text);
  await thread.post("Got it!");
});
```

### Chat SDK Version

4.16.0

### Node.js Version

20.x (Vercel serverless)

### Platform Adapter

Slack

### Operating System

macOS

### Additional Context

- Redis provider: Upstash (Vercel KV) — HTTP-based, higher latency than TCP Redis, which widens the race window
- The SDK's own documentation and Slack app manifest template recommend subscribing to both `app_mention` and `message.channels`
- Removing `message.channels` from event subscriptions is a workaround but breaks `onSubscribedMessage` for non-mention follow-up messages in threads
- The issue is more pronounced on warm serverless instances where both events arrive near-simultaneously; on cold starts, the latency difference sometimes lets `app_mention` win the race

haydenbleasel:
## Summary

Fixes #180

- Added `setIfNotExists` to the `StateAdapter` interface — an atomic set-if-not-exists operation backed by Redis `SET NX PX` and memory-based check-and-set
- Replaced the non-atomic GET-then-SET deduplication pattern in `Chat.handleIncomingMessage` with a single `setIfNotExists` call
- The old pattern allowed both Slack `app_mention` and `message.channels` events (same `ts`) to pass the dedupe check simultaneously, causing ~75% mention drop rate

## Test plan

- [ ] Verify `setIfNotExists` unit tests pass for `MemoryStateAdapter` (set, reject duplicate, TTL expiry)
- [ ] Verify Chat-level dedupe tests pass (atomic call check, concurrent race simulation, message-type mention detection)
- [ ] Deploy to staging with both `app_mention` and `message.channels` subscriptions and confirm mentions respond reliably

### DIFF
--- a/.changeset/bright-regions-dance.md
+++ b/.changeset/bright-regions-dance.md
@@ -1,0 +1,8 @@
+---
+"@chat-adapter/state-ioredis": patch
+"@chat-adapter/state-memory": patch
+"@chat-adapter/state-redis": patch
+"chat": patch
+---
+
+fix: non-atomic message deduplication causes app_mention events to be silently dropped


packages/state-ioredis/src/index.ts
```diff
@@ -217,6 +217,23 @@ export class IoRedisStateAdapter implements StateAdapter {
...
+    value: unknown,
+    ttlMs?: number
+  ): Promise<boolean> {
+    this.ensureConnected();
+
+    const cacheKey = this.key("cache", key);
+    const serialized = JSON.stringify(value);
+
+    const result = ttlMs
+      ? await this.client.set(cacheKey, serialized, "PX", ttlMs, "NX")
```
cramforce, line 231: Doesn't this override?

haydenbleasel: Had Claude take a look, apparently not. NX flag in Redis means "only set if the key doesn't exist." If key exists, command returns null and method returns false.

PR #193, opened 2026-03-07
feat: add forceReleaseLock + onLockConflict for interrupt/steerability

gakonst:
## Problem

When building AI agents on top of the chat SDK, users need to be able to interrupt a running agent by sending a follow-up message. Currently, if a handler is still processing (e.g., streaming an LLM response), the thread lock blocks the new message and it gets silently dropped via `LockError`.

This makes it impossible to steer or interrupt long-running agent turns from chat.

## Solution

Two changes:

1. **`StateAdapter.forceReleaseLock(threadId)`** — unconditionally releases a thread lock regardless of ownership token. The previous lock holder's `finally` block will call `releaseLock()` with a stale token, which is already a no-op (both Redis and memory impls verify the token before deleting).

2. **`onLockConflict` config option** — controls behavior when `acquireLock` fails:
   - `'drop'` (default, preserves current behavior) — throw `LockError`
   - `'force'` — force-release the existing lock and re-acquire
   - `(threadId, message) => 'force' | 'drop'` — callback for custom logic

```typescript
const chat = new Chat({
  userName: 'my-bot',
  state: createRedisState(),
  onLockConflict: 'force', // always allow interrupt
});
```

## Why this matters

Any chat SDK user building long-running handlers (AI agents, workflows, async tasks) needs this. The current lock-or-drop behavior was designed for short idempotent handlers, not multi-minute streaming sessions. This is a non-breaking, backwards-compatible addition.

### DIFF
--- a/apps/docs/content/docs/error-handling.mdx
+++ b/apps/docs/content/docs/error-handling.mdx
@@ -97,7 +97,7 @@ See the [feature matrix](/docs/adapters) for which features are supported on eac
 
 ### LockError
 
-Thrown when the SDK fails to acquire a distributed lock on a thread (used to prevent concurrent processing of messages in the same thread).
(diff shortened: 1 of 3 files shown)


haydenbleasel:
great stuff @gakonst 🙏

PR #205, opened 2026-03-09
Expose stream options on thread.post

notkadez:
## Summary
- add `thread.post(message, { stream: ... })` support for public stream options
- forward Slack stream presentation options like `taskDisplayMode` and `stopBlocks`
- update docs to use the public `thread.post(...)` API instead of undocumented `thread.stream(...)`

## Problem
- Docs say to use thread.stream, which is not exposed.
- There is no way to stream to slack using plan mode.


notkadez:
@cramforce - this is ready for review when you get a chance. Let me know if there is anything blocking.

cramforce:
@visyat Let me know if this is needed now that we have (but did we document?) PostableObject?

PR #230, opened 2026-03-12
feat(slack): resolve @displayname mentions to <@USER_ID> in outgoing messages

cramforce:
Build a reverse index from display name → user IDs during lookupUser(), track thread participants on incoming messages, and resolve @name mentions to Slack's <@USER_ID> format before sending. Disambiguates using thread participants when multiple users share a display name. Also increases user/channel cache TTLs to 8 days.

### DIFF
--- a/packages/adapter-slack/src/index.ts
+++ b/packages/adapter-slack/src/index.ts
@@ -244,6 +244,18 @@ interface SlackMemberJoinedChannelEvent {
   user: string;
 }
 
+/** Slack user_change event payload */
+interface SlackUserChangeEvent {
+  event_ts: string;
+  type: "user_change";
+  user: {
+    id: string;
+    name?: string;
+    real_name?: string;
+    profile?: { display_name?: string; real_name?: string };
+  };
+}
+
 /** Slack webhook payload envelope */
 interface SlackWebhookPayload {
   challenge?: string;
@@ -253,7 +265,8 @@ interface SlackWebhookPayload {
     | SlackAssistantThreadStartedEvent
     | SlackAssistantContextChangedEvent
     | SlackAppHomeOpenedEvent
-    | SlackMemberJoinedChannelEvent;
+    | SlackMemberJoinedChannelEvent
+    | SlackUserChangeEvent;
   event_id?: string;
   event_time?: number;
   team_id?: string;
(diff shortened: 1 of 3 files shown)


packages/adapter-slack/src/index.ts
```diff
@@ -1927,6 +1962,136 @@ export class SlackAdapter implements Adapter<SlackThreadId, unknown> {
...
+      );
+      participants = new Set(participantList);
+    }
+
+    // Replace mentions in text
+    return text.replace(mentionPattern, (match, name: string) => {
+      const idx = text.indexOf(match);
+      if (idx > 0 && text[idx - 1] === "<") {
+        return match;
+      }
```
matchai: In the event of two mentions in the same message, this `indexOf` will resolve to the first instance for every match. Patched up by using the offset:


```suggestion
    return text.replace(mentionPattern, (match, name: string, offset: number) => {
      if (offset > 0 && text[offset - 1] === "<") {
        return match;
      }
```

packages/adapter-slack/src/index.ts
```diff
@@ -725,6 +727,17 @@ export class SlackAdapter implements Adapter<SlackThreadId, unknown> {
             { displayName, realName },
             SlackAdapter.USER_CACHE_TTL_MS
           );
+
+        // Build reverse index: display name → user IDs (skip if already present)
```
matchai, line 744: When a user changes their name, the old mapping persists and the new one isn't created until the user cache expires.

Not a big deal unless users swap names with each other.

packages/adapter-slack/src/index.ts
```diff
@@ -357,8 +358,9 @@ export class SlackAdapter implements Adapter<SlackThreadId, unknown> {
   private _botUserId: string | null = null;
   private _botId: string | null = null; // Bot app ID (B_xxx) - different from user ID
   private readonly formatConverter = new SlackFormatConverter();
-  private static USER_CACHE_TTL_MS = 60 * 60 * 1000; // 1 hour
-  private static CHANNEL_CACHE_TTL_MS = 60 * 60 * 1000; // 1 hour
+  private static USER_CACHE_TTL_MS = 8 * 24 * 60 * 60 * 1000; // 8 days
```
matchai, line 374: The userId → displayName mapping will be stale here for 8 days. This'll cause cosmetic issues with the agent using an old user name after a name change for to up to 8 days.

We have `users:read` in the manifest so we should subscribe to `user_change` events for cache invalidation.

matchai:
@cramforce I've gone ahead and stacked #236 on your PR to implement cache invalidation 👍

packages/adapter-slack/src/index.ts
matchai: Scratch that. The stale reverse entry is harmless. Slack renders `<@USER_ID>` as their current name anyway.

PR #233, opened 2026-03-13
fix(slack): use replace offset for duplicate mentions, invalidate cache on user_change

matchai:
Fixes three issues found during review of #230. Stacks on #234.

### Duplicate mention bug

`resolveOutgoingMentions` used `text.indexOf(match)` inside the `.replace()` callback, which always returns the index of the *first* occurrence. When the same `@name` appears multiple times, subsequent matches get their `<` prefix check evaluated against the wrong position. Uses the `offset` parameter that `.replace()` already provides.

### Stale user cache on name changes

The user cache (`slack:user:{userId}`) has an 8-day TTL. When a Slack user changes their display name, the cache serves stale data until expiry — incoming messages show the old name and the reverse index never learns the new name.

Subscribes to Slack's `user_change` event and:
1. Reads the old cached display name
2. Removes the user ID from the old name's reverse index (via `removeFromList` from #234)
3. Deletes the user cache entry

The next `lookupUser` call hits the Slack API and picks up the new profile. Requires adding `user_change` to `bot_events` in the Slack app manifest (`users:read` scope is already present).

### Test comment

Fixed misleading "fire-and-forget" comment — the code actually `await`s participant tracking.

### DIFF
--- a/.changeset/fix-mention-offset-user-change.md
+++ b/.changeset/fix-mention-offset-user-change.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": patch
+---
+
+Fix duplicate mention resolution using replace offset, invalidate user cache on profile changes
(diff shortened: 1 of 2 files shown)


matchai:
Superseded by combined PR

PR #234, opened 2026-03-13
feat(state): add removeFromList to StateAdapter

matchai:
Adds `removeFromList(key, value)` to the `StateAdapter` interface, completing the list CRUD (`appendToList`, `getList`, `removeFromList`).

Removes the first occurrence of `value` from the list at `key`. Returns `true` if an element was removed, `false` if not found.

### Implementations

- **memory**: `JSON.stringify` comparison + splice
- **redis**: Lua `LREM key 1 value`
- **ioredis**: Lua `LREM key 1 value`
- **pg**: `DELETE ... WHERE ctid = (SELECT ctid ... LIMIT 1)`
- **mock adapters**: same as memory

Stacks on #230.

### DIFF
--- a/.changeset/add-remove-from-list.md
+++ b/.changeset/add-remove-from-list.md
@@ -1,0 +1,9 @@
+---
+"chat": minor
+"@chat-adapter/state-memory": minor
+"@chat-adapter/state-redis": minor
+"@chat-adapter/state-ioredis": minor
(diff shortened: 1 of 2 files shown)


matchai:
Superseded by combined PR — removeFromList no longer needed for this fix

PR #235, opened 2026-03-13
Expose thread.toJSON() on the public Thread type

0xd669:
## Summary
- add `toJSON(): SerializedThread` to the public `Thread` interface
- add a consumer-facing type regression test for `import type { Thread } from "chat"`
- include a patch changeset for the public type fix

## Problem
`ThreadImpl` already exposes `toJSON()` at runtime and the docs/examples use `thread.toJSON()`, but the public `Thread` type in `dist/index.d.ts` did not include that method. This caused consumer builds to fail with:

```ts
Property 'toJSON' does not exist on type 'Thread<...>'
```

## Testing
- `pnpm validate`

### DIFF
--- a/.changeset/green-tables-lie.md
+++ b/.changeset/green-tables-lie.md
@@ -1,0 +1,5 @@
+---
+"chat": patch
+---
+
+Expose thread.toJSON() on the public Thread type so generated declarations match the runtime API and docs.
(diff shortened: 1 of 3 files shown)


haydenbleasel:
Hey @jacehwang, thanks for catching this and putting up the fix! 🙌

This has actually already been fixed on `main` — `toJSON(): SerializedThread` is already on the public `Thread` interface (line 955 in `types.ts`).

The type regression test is a nice idea though! If we wanted to add something like that, we'd probably lean toward vitest's `expectTypeOf`/`assertType` rather than shelling out to `tsc` with a temp project — lighter weight and less fragile. But appreciate the thought!

Closing this one out since the fix is already in. Thanks again!

PR #245, opened 2026-03-15
fix(telegram): set parse_mode for markdown messages

### LINKED ISSUE #226: Telegram adapter: parse_mode not set for markdown messages
## Bug

When sending messages with `{ markdown: ... }` payload, the Telegram adapter does not set `parse_mode` on the Telegram API call. This causes Telegram to render the message as plain text, showing raw markdown characters (`**bold**`, `_italic_`) instead of formatted text.

## Root Cause

In `@chat-adapter/telegram`, both `postMessage` and `editMessage` only set `parseMode` when a `card` is present:

```js
const parseMode = card ? TELEGRAM_MARKDOWN_PARSE_MODE : void 0;
```

When the message payload is `{ markdown: "**hello**" }`, the `formatConverter.renderPostable()` correctly processes the markdown text, but `parseMode` remains `undefined` because there's no card. The Telegram API then treats the text as plain.

## Expected Behavior

`parse_mode` should be set to `"Markdown"` when the message contains a `markdown` field:

```js
const parseMode = (card || (typeof message === "object" && message !== null && "markdown" in message))
  ? TELEGRAM_MARKDOWN_PARSE_MODE
  : void 0;
```

## Affected Methods

- `postMessage` (line ~527 in dist)
- `editMessage` (line ~593 in dist)

## Reproduction

```ts
const chat = new Chat({ adapters: { telegram: telegramAdapter } });
const thread = await chat.getThread("telegram", threadId);

// This renders as plain text with visible ** characters
await thread.post({ markdown: "**bold** and _italic_" });

// This works correctly (has card → parseMode is set)
await thread.post({ card: Card({ children: [CardText("hello")] }), fallbackText: "hello" });
```

## Workaround

We're using `bun patch` to fix this locally:

```diff
-    const parseMode = card ? TELEGRAM_MARKDOWN_PARSE_MODE : void 0;
+    const parseMode = (card || (typeof message === "object" && message !== null && "markdown" in message)) ? TELEGRAM_MARKDOWN_PARSE_MODE : void 0;
```

## Version

`@chat-adapter/telegram@4.20.0`

bensabic:
`postMessage` and `editMessage` only set `parse_mode` when a card was present, causing `{ markdown: "..." }` messages to render as plain text with visible formatting characters. Now also sets `parse_mode` to `"Markdown"` when the message has a `markdown` field.

Closes #226

### DIFF
--- a/packages/adapter-telegram/src/index.ts
+++ b/packages/adapter-telegram/src/index.ts
@@ -556,10 +556,7 @@ export class TelegramAdapter
 
     const card = extractCard(message);
     const replyMarkup = card ? cardToTelegramInlineKeyboard(card) : undefined;
-    const hasMarkdown =
-      typeof message === "object" && message !== null && "markdown" in message;
-    const parseMode =
-      card || hasMarkdown ? TELEGRAM_MARKDOWN_PARSE_MODE : undefined;
+    const parseMode = this.resolveParseMode(message, card);
     const text = this.truncateMessage(
       convertEmojiPlaceholders(
         card
@@ -645,10 +642,7 @@ export class TelegramAdapter
 
     const card = extractCard(message);
     const replyMarkup = card ? cardToTelegramInlineKeyboard(card) : undefined;
-    const hasMarkdown =
-      typeof message === "object" && message !== null && "markdown" in message;
-    const parseMode =
-      card || hasMarkdown ? TELEGRAM_MARKDOWN_PARSE_MODE : undefined;
+    const parseMode = this.resolveParseMode(message, card);
     const text = this.truncateMessage(
       convertEmojiPlaceholders(
         card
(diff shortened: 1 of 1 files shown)


packages/adapter-telegram/src/index.ts
```diff
@@ -556,7 +556,10 @@ export class TelegramAdapter
 
     const card = extractCard(message);
     const replyMarkup = card ? cardToTelegramInlineKeyboard(card) : undefined;
-    const parseMode = card ? TELEGRAM_MARKDOWN_PARSE_MODE : undefined;
+    const hasMarkdown =
```
haydenbleasel: @bensabic This logic is duplicated below, can pull out into a function for cleanliness 👍

PR #256, opened 2026-03-16
fix(discord): remove duplicate text when posting card messages

### LINKED ISSUE #246: Discord is doing text + embed when you use card components
### Bug Description

Image

### Steps to Reproduce

https://chat-sdk.dev/docs/cards

### Expected Behavior

only cards to be there, not extra text duplicating the same information

### Actual Behavior

cards + text was there

### Code Sample

```typescript

```

### Chat SDK Version

4.20.1

### Node.js Version

_No response_

### Platform Adapter

_No response_

### Operating System

None

### Additional Context

_No response_

haydenbleasel:
Resolves #246

### DIFF
--- a/.changeset/fix-discord-duplicate-card-content.md
+++ b/.changeset/fix-discord-duplicate-card-content.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/discord": patch
+---
+
+Fix duplicate content display when sending card messages on Discord
--- a/packages/adapter-discord/src/index.test.ts
+++ b/packages/adapter-discord/src/index.test.ts
@@ -1536,6 +1536,42 @@ describe("postMessage", () => {
 
     spy.mockRestore();
   });
+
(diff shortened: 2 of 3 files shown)


bensabic (approved):

bensabic (dismissed):
Thanks for fixing the duplicate content issue on create paths. I think there’s still one gap on `editMessage`.

In `packages/adapter-discord/src/index.ts`, the card branch now omits `payload.content` entirely when PATCHing an existing message. Discord’s Edit Message API treats fields as optional and nullable, so leaving `content` out does not reliably clear the previous body. If a normal text message is edited into a card, the old text can still remain alongside the embed/components.

I think this should explicitly clear `content` in the edit path (`""` or `null`, depending on what we want to standardize on), and the test should cover editing a non-empty text message into a card rather than only asserting that the outgoing payload omits `content`.

dancer (approved):

PR #257, opened 2026-03-16
fix(chat): export standalone reviver for workflow-safe deserialization

### LINKED ISSUE #243: Durable chat sessions guide fails with workflow-node-module-error
Following the "Durable chat sessions with Next.js, Workflow, and Redis" guide as written causes `pnpm dev` to fail during workflow bundling.

Versions:
- chat: 4.20.1
- @chat-adapter/slack: 4.20.1
- @chat-adapter/state-redis: 4.20.1
- workflow: 4.2.0-beta.70
- next: 16.1.6

The guide imports `bot` into `workflows/durable-chat-session.ts` and uses `bot.reviver()` inside the workflow. With the versions above, that pulls `@chat-adapter/slack` into the workflow bundle and fails with:

```
✘ [ERROR] You are attempting to use "@chat-adapter/slack" which depends on Node.js modules. Packages that depend on Node.js modules are not available in workflow functions.

lib/bot.ts:5:9:
  5 │   slack: createSlackAdapter(),
    │          ~~~~~~~~~~~~~~~~~~
    ╵          Move this function into a step function.
```

Expected:
- The docs example should work as written, or
- The guide should mention that `bot` cannot be imported into the workflow module and should only be used inside `"use step"` functions.

Workaround:
Keep the workflow on serialized thread/message data and only import `@/lib/bot` inside step functions.

haydenbleasel:
## Summary

- Adds a standalone `reviver` function exported from `chat` that deserializes Thread/Channel/Message objects without requiring a Chat instance (and its adapter Node.js dependencies)
- Updates `Chat.reviver()` to delegate to the standalone function internally
- Updates the durable chat sessions guide to use `reviver` from `chat` and dynamically import `bot` inside `"use step"` functions

Closes #243

## Test plan

- [x] `pnpm validate` passes (knip, lint, typecheck, tests, build)
- [x] New tests for standalone `reviver` in `serialization.test.ts`
- [ ] Verify the updated guide pattern works with a real Vercel Workflow project

### DIFF
--- a/.changeset/six-dolls-smash.md
+++ b/.changeset/six-dolls-smash.md
@@ -1,0 +1,5 @@
+---
+"chat": minor
+---
+
+export standalone reviver for workflow-safe deserialization without adapter dependencies
--- a/apps/docs/content/docs/api/chat.mdx
+++ b/apps/docs/content/docs/api/chat.mdx
@@ -475,3 +475,13 @@ Get a `JSON.parse` reviver that deserializes `Thread` and `Message` objects from
 const data = JSON.parse(payload, bot.reviver());
 await data.thread.post("Hello from workflow!");
 ```
+
+There is also a standalone `reviver` export that works without a `Chat` instance. This is useful in Vercel Workflow functions where importing the full Chat instance (with its adapter dependencies) is not possible:
+
+```typescript
+import { reviver } from "chat";
+
+const data = JSON.parse(payload, reviver) as { thread: Thread; message: Message };
+```
+
+The standalone reviver uses lazy adapter resolution - the adapter is looked up from the Chat singleton when first accessed. Call `chat.registerSingleton()` before using thread methods like `post()` (typically inside a `"use step"` function).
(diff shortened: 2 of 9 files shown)


bensabic (changes requested):
Thanks for working through the Workflow sandbox issue, @haydenbleasel! The standalone `reviver` direction makes sense, but I don’t think this is quite ready to merge yet.

I found two follow-ups that still need to be addressed:

1. `ChannelImpl` has the same lazy round-trip bug as `ThreadImpl`. A `chat:Channel` revived via the new standalone `reviver` still can’t be serialized again without a registered singleton, because `toJSON()` derives `adapterName` from `this.adapter.name` instead of preserving the stored adapter name. This makes the new helper incomplete for channel payloads too.

2. The docs are now internally inconsistent around the new API. The durable session guide points readers to the Chat API for `reviver`, but `apps/docs/content/docs/api/chat.mdx` still only documents `bot.reviver()`, not the new root export.

If you fix the lazy serialization path by preserving `adapterName` on round-trip and update the API docs to cover the new export, this should be in a better shape.

bensabic:
@dancer hey Josh, do you have time to review and edit this PR?

bensabic (approved):

PR #269, opened 2026-03-19
Support for Telegram attachments

alizahid:
## Summary

Sets up attachment support for Telegram. This allows us to send certain files (primarily images) to be send as attachments so they show in Telegram chat as proper images instead of file uploads.

I've added support for audio files, other file, images, and videos as attachments. Can easily extend with other types in the future like voice messages and location, etc.

## Test plan

Patched my local with this PR and sent images, audio and other files.

Also added tests for `extractAttachments`.


dancer:
#485

PR #283, opened 2026-03-22
Feat/adapter teams sdk

kushalbhabra:


bensabic:
Thanks for the PR! Since we already have a [Microsoft Teams adapter](https://chat-sdk.dev/adapters/teams), we'll have to close this PR.

kushalbhabra:
Hi @bensabic 


Thanks for the review and the overall chat SDK.

As I understand, the current teams adapter uses only a very small subset of features of teams as it doesn't use teams.ts SDK, instead uses the underlying bot framework library.

Few important flows include streaming, user sso and auth flows, message extensions and link unfurling etc. are missing from the current adapter.

I understand teams.ts uses a express server which is not compatible directly with chat SDK. This PR is an attempt to bridge that gap using a teams.ts router component.

Would like to know if there's active work going on for making the current teams adapter to provide full support for all teams features.

PR #286, opened 2026-03-23
fix(chat): use adapter to parse channel id

yjl9903:
## Summary

The value of `thread.channelId` is wrong, it should use `adapter.channelIdFromThreadId(threadId)` method provided by adapter to generate a channel id instead of hard-code this logic for all providers.

## Test plan

Fix related testcase asserts in the current codebase.

### DIFF
--- a/.changeset/thin-sloths-joke.md
+++ b/.changeset/thin-sloths-joke.md
@@ -1,0 +1,5 @@
+---
+"chat": patch
+---
+
+Use adapter to parse channel id for `thread.channelId`


yjl9903:
@cramforce @haydenbleasel Hello, could you please help review this PR when you have a chance? Thanks a lot!

For now, the behavior of `thread.channelId` is strange and does not align with `thread.channel.id`.

cramforce:
Thanks! Would you mind running pnpm changeset?

yjl9903:
> Thanks! Would you mind running pnpm changeset?

Done.

PR #293, opened 2026-03-24
Slack: upgrade streamed markdown tables to native blocks

mohankumarelec:
## What changed
- upgrades streamed markdown tables into native Slack table blocks once stream is done
- keeps `stopBlocks` when doing that final `chat.update`
- adds tests for table + no-table stream paths and postMessage behavior

## Validation
- `pnpm check`
- `pnpm test --filter=@chat-adapter/slack...`

Small behavior fix, should be safer for current bots too.

### DIFF
--- a/.changeset/fuzzy-tigers-share.md
+++ b/.changeset/fuzzy-tigers-share.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": patch
+---
+
+Upgrade streamed markdown tables to native Slack table blocks after stream completion, and preserve stop blocks during the final message update.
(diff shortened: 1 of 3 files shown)


cramforce:
I'm not sure this is desirable. It's a tough one

mohankumarelec:
@cramforce , since slack now supports native markdown table, it has few extra features like rendering large tables in an expanded window, exporting the table in tsv and so on. 

Do you see downsides of implementing this ?

cramforce:
The main downside is that the shape of the rendering ends at the very end of the stream. So, if you are actually following along you have a janky experience.

In fact, we originally did this and it was often jarring. If we knew that nobody was watching, it would definitely be better to use native rendering.

mohankumarelec:
@cramforce , got it , makes sense

PR #302, opened 2026-03-26
Migrate from BotFramework to TeamsSDK

heyitsaamir:
Hello! I am the maintainer for [TeamsSDK](https://microsoft.github.io/teams-sdk/welcome), which is the official SDK for Microsoft Teams. We noticed that y'all are using BotFramework in the adapter here which is actually now deprecated. We took the initiative to migrate you guys over to TeamsSDK. Hope that's okay!

Some wins:
* Port to our SDK which is now being actively maintained.
* Better type safety all around - we have better type safety not only for Teams entities, but also Graph so the adapter doesn't need to maintain Graph types).
* Less code for setup 
* Reactions support

Pending:
* I need to go through all the unit tests that y'all had and make sure I'm not missing any major ones. I manually tested everything so it should be okay, but will add UTs as well.

What didn't make it:
* We don't really support Certificates. We have client-secret support, User managed identity support, and Federated Identity Credentials support. But no certs. If someone really wanted it, then can pass their own token generator as part of `token` option (which is basically a factory that generates a token). Our general logic was that if someone is using certs, they might as well just use secrets. 
* I changed the AdapterOptions to be closer to what TeamsSDK App has instead of what BotFrameworked has. If we want to maintain backward compatibility, I'm happy to keep that outward facing interface and map to our App options. 

Some things I would like to add in a followup (or even in this):
* Simpler code - 1 giant file is hard for me and Claude to work through :) Wanted an okay from someone on the team here before I move some code to their own smaller files.
* I would like to migrate the cards logic over to our cards package as well which also offers better type safety over just raw jsons.
* We have modal support as well. I'd like to add support for that next.
* I know the setup is inconvenient right now, so we are working on a solution improve that as well.

## Test plan

Tested the scenarios in the example bot.
Eg Channels: 
image

DMs:
image

(Happy to share more screenshots / video if needed).

### DIFF
--- a/packages/adapter-teams/src/graph-api.ts
+++ b/packages/adapter-teams/src/graph-api.ts
@@ -92,36 +92,14 @@ export class TeamsGraphReader {
       let hasMoreMessages = false;
 
       if (direction === "forward") {
-        const allMessages: GraphMessage[] = [];
-
-        const firstPage = await this.deps.graph.call(chats.messages.list, {
+        const response = await this.deps.graph.call(chats.messages.list, {
           "chat-id": baseConversationId,
-          $top: 50,
-          $orderby: ["createdDateTime desc"],
+          $top: limit,
+          $orderby: ["createdDateTime asc"],
+          $filter: cursor ? `createdDateTime gt ${cursor}` : undefined,
         });
-        allMessages.push(...(firstPage.value || []));
-
-        let nextLink = firstPage["@odata.nextLink"] ?? undefined;
-        while (nextLink) {
-          const page = await this.graphGetNextLink<typeof firstPage>(nextLink);
-          allMessages.push(...(page.value || []));
-          nextLink = page["@odata.nextLink"] ?? undefined;
-        }
-
-        allMessages.reverse();
-
-        let startIndex = 0;
-        if (cursor) {
-          startIndex = allMessages.findIndex(
-            (msg) => msg.createdDateTime && msg.createdDateTime > cursor
-          );
-          if (startIndex === -1) {
-            startIndex = allMessages.length;
-          }
-        }
-
-        hasMoreMessages = startIndex + limit < allMessages.length;
-        graphMessages = allMessages.slice(startIndex, startIndex + limit);
+        graphMessages = (response.value || []) as GraphMessage[];
+        hasMoreMessages = graphMessages.length >= limit;
       } else {
         const response = await this.deps.graph.call(chats.messages.list, {
           "chat-id": baseConversationId,
@@ -262,36 +240,16 @@ export class TeamsGraphReader {
           hasMoreMessages = graphMessages.length >= limit;
         }
       } else if (direction === "forward") {
-        const allMessages: GraphMessage[] = [];
-        const firstPage = await this.deps.graph.call(chats.messages.list, {
+        const response = await this.deps.graph.call(chats.messages.list, {
           "chat-id": baseConversationId,
-          $top: 50,
-          $orderby: ["createdDateTime desc"],
+          $top: limit,
+          $orderby: ["createdDateTime asc"],
+          $filter: options.cursor
+            ? `createdDateTime gt ${options.cursor}`
+            : undefined,
         });
-        allMessages.push(...(firstPage.value || []));
-        let nextLink = firstPage["@odata.nextLink"] ?? undefined;
-        while (nextLink) {
-          const page = await this.graphGetNextLink<{
-            value: GraphMessage[];
-            "@odata.nextLink"?: string;
-          }>(nextLink);
-          allMessages.push(...(page.value || []));
-          nextLink = page["@odata.nextLink"] ?? undefined;
-        }
-
-        allMessages.reverse();
-        let startIndex = 0;
-        if (options.cursor) {
-          const cursorVal = options.cursor;
-          startIndex = allMessages.findIndex(
-            (msg) => msg.createdDateTime && msg.createdDateTime > cursorVal
-          );
-          if (startIndex === -1) {
-            startIndex = allMessages.length;
-          }
-        }
-        hasMoreMessages = startIndex + limit < allMessages.length;
-        graphMessages = allMessages.slice(startIndex, startIndex + limit);
+        graphMessages = (response.value || []) as GraphMessage[];
+        hasMoreMessages = graphMessages.length >= limit;
       } else {
         const response = await this.deps.graph.call(chats.messages.list, {
           "chat-id": baseConversationId,


bensabic:
Hey @heyitsaamir!

Thank you so much for this.

Quick comment: would you be able to also add support for Task Modules?

heyitsaamir:
> Hey @heyitsaamir!
> 
> 
> 
> Thank you so much for this.
> 
> 
> 
> Quick comment: would you be able to also add support for Task Modules?

Yes! Would you like me to add that in this same pr or follow up? My plan was on a follow up.

heyitsaamir:
@bensabic Created this Issue when it comes to modals on Teams - https://github.com/vercel/chat/issues/306. Lmk what you think!

bensabic:
@heyitsaamir I wasn't sure what they meant either, looked like spam or a bot—just deleted the comments to reduce noise.

Thank you for the very detailed issue, I just replied to it for the sake of the PR thread being cleaner. It makes sense why modals support would need to be in a separate PR, since the underlying mechanics itself for how we handle modals would have to be tweaked slightly.

Unrelated: does the Teams SDK have support for slash commands? That would be amazing if so. As the dream is for all adapters to be as close together as possible in terms of functionality.

heyitsaamir:
> @heyitsaamir I wasn't sure what they meant either, looked like spam or a bot—just deleted the comments to reduce noise.
> 
> Thank you for the very detailed issue, I just replied to it for the sake of the PR thread being cleaner. It makes sense why modals support would need to be in a separate PR, since the underlying mechanics itself for how we handle modals would have to be tweaked slightly.
> 
> Unrelated: does the Teams SDK have support for slash commands? That would be amazing if so. As the dream is for all adapters to be as close together as possible in terms of functionality.

Sweet. I'm happy to put up a PR stacked on this, or once this goes in (if y'all are okay with it).

As for slash commands - we _currently_ have a version of it:
https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/create-a-bot-commands-menu?tabs=desktop%2Cjavascript
Here, it will _suggest_ some slash commands for the user, and you declare them in your manifest. But triggering them essentially just dumps the command in your compose box. And the user needs to send them. So it acts like a regular, pre-built, message. 
There's actually something called "[suggested actions](https://microsoft.github.io/teams-sdk/csharp/in-depth-guides/ai/best-practices?_highlight=sugg#suggested-actions)" which is basically this, but dyanamic (i.e. you can pass it alongside your message). The user can click on these suggestions and it'll do the same -- put it in the compose box that the user then needs to send.

The biggest downside with the current method is that in a group-chat, everyone sees the slash-message. In a DM it doesn't really matter that much.

I will say that there are some more things coming soon that will enable better slash-command like experience soon.

heyitsaamir:
btw @bensabic I refactored a bit to split the giant index.ts into two files. Hope that's okay. The first file was just getting wildly unwieldy.

bensabic:
> 
> Sweet. I'm happy to put up a PR stacked on this, or once this goes in (if y'all are okay with it).
> 
> As for slash commands - we _currently_ have a version of it: https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/create-a-bot-commands-menu?tabs=desktop%2Cjavascript Here, it will _suggest_ some slash commands for the user, and you declare them in your manifest. But triggering them essentially just dumps the command in your compose box. And the user needs to send them. So it acts like a regular, pre-built, message. There's actually something called "[suggested actions](https://microsoft.github.io/teams-sdk/csharp/in-depth-guides/ai/best-practices?_highlight=sugg#suggested-actions)" which is basically this, but dyanamic (i.e. you can pass it alongside your message). The user can click on these suggestions and it'll do the same -- put it in the compose box that the user then needs to send.
> 
> The biggest downside with the current method is that in a group-chat, everyone sees the slash-message. In a DM it doesn't really matter that much.
> 
> I will say that there are some more things coming soon that will enable better slash-command like experience soon.

Re: Modals, I'd say wait until this PR goes in, then we can cross that bridge when we come to it.

Re: Slash commands, I'd ideally rather wait to add support for that until the Teams SDK has slash commands which are closer to Slack functionality. Mainly so the UX is more familiar to users and consistent across platforms.

heyitsaamir:
Ah thanks for the closer look! Yeah I called out the breaking options in the description, so I'm happy to build a translation layer between the two. 

Let me go through these suggestions, and get back to you :). Thanks Ben.

cramforce:
This is amazing. I'm doing some manual testing

cramforce:
Just naively booting this I'm getting `error: Error [NetworkError]: Teams API error during startTyping: app not started`

@heyitsaamir Do you have a lead what to do?

heyitsaamir:
> Just naively booting this I'm getting `error: Error [NetworkError]: Teams API error during startTyping: app not started`
> 
> @heyitsaamir Do you have a lead what to do?

Oh huh, that shouldn't happen in 2.0.6 (it was the case before that though). It's using 2.0.6 right? I'll do some digging too to make sure I didn't assume anything.

Oh, @cramforce I did change the expectations of the .env file (which i'm fixing, that could be one reason? -- see Readme?)

heyitsaamir:
@cramforce pushed up a change to respect previous public facing adapter config. So try now?

The main thing that we currently don't support is Certs. Our logic was that if folks want to use certs, they might as well use secrets. If we really want to support certs, we could either build a custom token-getter. Or consider adding it in the sdk itse.f.

bensabic:
@heyitsaamir just curious why reaction support had to be removed?

heyitsaamir:
Hello @bensabic! I have some updates:

1. Breaking config contract — This is reverted back to what was originally supported. The only thing this still breaks is cert support. We removed certs support because if someone is using certs, they might as well use secrets. If y'all really want this, we can look into including it in the sdk, or passing in a `token` generator as part of the App options.

2. Typecheck / test failures — These should all work now. I pulled back all the tests too.

3. Channel-thread history regression — We have an endpoint to actually look teams aad group ids up for a givem team id. So I added support for that and cached it. Our APIs are sort of strange because _some_ events have the aadGroupId, and _some_ don't. I asked our service team to tell us the logic here and also provide this id always. Let's see what happens!

4. WebhookOptions concurrency bug — I save the webhook options in the bridge now and cache it for an activity id (unique per activity).

5. Reaction support — Checked with the feature team here, and this is actually not fully rolled out. (I think a few weeks out). I'm happy to add this back in when it's ready.

6. Related ^, we're also getting ephemeral messages soon as well, so that'll be nice for those scenarios.

7. Tenant autodetection in proactive paths — Falling back to appConfig.tenantId is probably okay for most common scenarios, but using the cached tenant id is def recommended. Anyway, the config should now beback to what it was. 

Had a question for you -- why is service url cached? It shouldn't really be needed (esp for proactive scenarios). Any idea? The service url is generally just useful for different kinds of environments (government etc.), but it can just be set when constructing `App` instead of needing to cache it. (Not a biggie though, and can be changed later if y'all decide)

cramforce:
I just redeployed the adapter and it is working now during manual testing!

i don't recall why I was caching the serviceUrl. When I originally wrote/prompted the code, I was just pushing through the non-great docs to somehow get it to work. I'd love it if you guys take on maintenance of the adapter as you  know best!

heyitsaamir:
@cramforce  we're interested in helping out :). We're trying to also improve our docs and experience in general so the feedback loop is mutually beneficial!

bensabic:
Hey @heyitsaamir, thank you again for offering to help and taking on this migration as well.

I pushed a commit with a few changes:

- **README**: Was out of sync with the code after the config revert commits — updated to use `appId`/`appPassword`/`appTenantId` and `TEAMS_APP_*` env vars. Also fixed reactions table.
- **`errors.ts`**: 403 was caught by the `AuthenticationError` block, making the `PermissionError` check dead code. Split them so 401 → `AuthenticationError`, 403 → `PermissionError`.
- **`index.ts`**: `NotImplementedError` args were swapped in `addReaction`/`removeReaction` vs. the convention used elsewhere.
- **Tests**: Removed `createMockBotAdapter`/`MockBotAdapter` compat aliases, renamed to `createMockTeamsApp`/`MockTeamsApp` throughout.
- Added changeset (minor bump).

Non-blocking nits for awareness:
- `graph-api.ts:691` — `@ts-expect-error` on `graph.http` for nextLink pagination is fragile if SDK internals change.
- Forward pagination fetches all pages before slicing — could be expensive for large channels.

All 123 tests pass and the package builds cleanly. Nice work on the module decomposition (types, errors, graph-api, bridge-adapter), much more maintainable than a single giant file.

heyitsaamir:
Thanks @bensabic. Already have a few of the fixes from what you mentioned propagating to the sdk. Will have some minor follow ups :). Thank you!

bensabic:
Sounds good @heyitsaamir! We're good to proceed with merging this once you're able to toss those follow-ups in.

heyitsaamir:
> Sounds good @heyitsaamir! We're good to proceed with merging this once you're able to toss those follow-ups in.

updated w/ the graph fixes. the sdk typing will most likely land by next week with a release. happy to bump the package then or wait. whatever you prefer!

bensabic:
@heyitsaamir  thanks for jumping on that so quickly! I'll merge it now. That's alright, I'm happy for it to be added via a patch later on. It can be bundled with other changes (e.g. new features).

bart-sofomo:
@heyitsaamir just a question about cards because right now they are not supporting `Input.ChoiceSet` -> do you plan to redesign to it uses it, etc.?

heyitsaamir:
> @heyitsaamir just a question about cards because right now they are not supporting `Input.ChoiceSet` -> do you plan to redesign to it uses it, etc.?
> 
> 

Good question. I think we can fix it. I'll take a look today

bart-sofomo:
> > @heyitsaamir just a question about cards because right now they are not supporting `Input.ChoiceSet` -> do you plan to redesign to it uses it, etc.?
> 
> Good question. I think we can fix it. I'll take a look today

That would be awesome - I think it's the last thing I'm waiting for.
Well you mentions few other updates but this one is kind of blocker for me.

PR #303, opened 2026-03-26
Fix Redis state typings for url and client options

haydenbleasel:
## Summary

- fix `createRedisState` typings so `url` appears in IntelliSense for `createRedisState({ url: "..." })`
- add an explicit factory options type and keep `REDIS_URL` env fallback
- support passing an existing Redis client and avoid closing caller-owned clients
- add tests for explicit `url` usage and existing-client usage
- include a patch changeset for `@chat-adapter/state-redis`

## Test plan

- Ran `pnpm build` and `pnpm test`, both successful

### DIFF
--- a/packages/state-redis/src/index.test.ts
+++ b/packages/state-redis/src/index.test.ts
@@ -118,6 +118,48 @@ describe("RedisStateAdapter", () => {
 
     await connectPromise;
     expect(resolved).toBe(true);
+    expect(client.connect).not.toHaveBeenCalled();
+  });
+
+  it("should wait for an injected client to become ready again after reconnecting", async () => {
+    const emitter = new EventEmitter();
+    const client = Object.assign(emitter, {
+      close: vi.fn(),
+      connect: vi.fn(),
+      isOpen: true,
+      isReady: true,
+    }) as unknown as RedisClientType & {
+      isOpen: boolean;
+      isReady: boolean;
+    };
+
+    const adapter = createRedisState({
+      client,
+      logger: mockLogger,
+    });
+
+    await adapter.connect();
+    expect(client.connect).not.toHaveBeenCalled();
+
+    client.isReady = false;
+    client.isOpen = false;
+    emitter.emit("reconnecting");
+
+    let resolved = false;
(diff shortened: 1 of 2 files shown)


haydenbleasel:
@bensabic fixed

bensabic:
@haydenbleasel Thanks for the follow-ups. The latest memory-state test change looks fine, and the original `isOpen` vs `isReady` issue is much closer, but I still think there’s one remaining bug in the injected Redis client path.

In `createRedisState({ client })`, `waitForReady()` currently rejects on the first `error` event. With node-redis, an open client can emit transient `error` events while reconnecting and later recover with `ready`, so `await adapter.connect()` can still fail even though the client would shortly become usable. I think this should keep waiting while the client is still open/reconnecting, and only reject on a terminal failure or close.

For example:

```ts
const client = createClient({ url: process.env.REDIS_URL! });
await client.connect();

const state = createRedisState({ client });

// Later, the connection drops briefly and node-redis starts reconnecting.
// During that period, it can emit `error` before it emits `ready` again.

await state.connect(); // should wait until the client is ready again
```

With the current implementation, if the injected client emits a transient error while reconnecting, `state.connect()` can reject immediately instead of waiting for the subsequent `ready` event. That makes the injected-client path fail spuriously even though the Redis client would have recovered on its own.

haydenbleasel:
Quick update on the last two commits:

- `7296aa0` fixes the injected Redis client wait path so `createRedisState({ client })` no longer rejects on the first transient `error` while the node-redis client is recovering. We now keep waiting for `ready` and only reject if the client actually ends before becoming ready.
- `9ad08ac` fixes the later reconnect case for injected clients. The adapter now tracks `ready` / `reconnecting` / `end` to keep its internal connected state in sync, so a previously connected injected client won’t short-circuit `state.connect()` during a reconnect. In that path we also avoid calling `client.connect()` for injected clients and just wait for the externally owned client to become ready again.

I also added regressions for both cases and re-ran:
- `pnpm --filter @chat-adapter/state-redis test`
- `pnpm --filter @chat-adapter/state-redis typecheck`
- `pnpm check`

All passed locally.

bensabic (approved):

Issue #231, opened 2026-03-13
Linear adapter: refreshClientCredentialsToken() omits app:mentionable scope


Fouadktb:
### Bug Description

The `LinearAdapter.refreshClientCredentialsToken()` method hardcodes the OAuth scope to `read,write,comments:create,issues:create` — it does not include `app:mentionable`.

### Steps to Reproduce

1. Create a Linear OAuth app with client credentials enabled and agent session events enabled
2. Install the app to a workspace with actor=app and scope=read,write,comments:create,issues:create,app:mentionable
3. Confirm the app shows as "Agent" in Linear and appears in @mention autocomplete
4. Start your server — the adapter calls refreshClientCredentialsToken() during initialize()
5. Check Linear — the app has switched from "Agent" to "App" and is no longer @mentionable

### Expected Behavior

The app should remain as "Agent" with app:mentionable capabilities after adapter initialization. The client credentials token request should include all scopes the app was installed with.

### Actual Behavior

The adapter requests a client credentials token with only read,write,comments:create,issues:create (hardcoded on https://github.com/vercel/chat/blob/main/packages/adapter-linear/src/index.ts), omitting app:mentionable. This downgrades the app from Agent to App on every server restart/token refresh.

Additionally, AgentSessionEvent webhooks (required for app:mentionable apps) are silently ignored — the adapter only handles Comment and Reaction types.

### Code Sample

```typescript
Workaround — pre-fetch token with correct scopes and bypass the adapter's built-in token request:

  const res = await fetch('https://api.linear.app/oauth/token', {
    method: 'POST',
    headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
    body: new URLSearchParams({
      grant_type: 'client_credentials',
      client_id: process.env.LINEAR_CLIENT_ID!,
      client_secret: process.env.LINEAR_CLIENT_SECRET!,
      scope: 'read,write,comments:create,issues:create,app:mentionable',
    }),
  })
  const { access_token } = await res.json()

  const bot = new Chat({
    adapters: {
      linear: createLinearAdapter({ accessToken: access_token }),
    },
  })
```

### Chat SDK Version

4.20.1

### Platform Adapter

Linear

### Operating System

macOS

### Additional Context

The adapter also silently ignores `AgentSessionEvent` webhooks (required for `app:mentionable` apps). First-class support for Linear Agent Sessions (emitting activities via `agentActivityCreate`) would be valuable.

PR #344, opened 2026-04-07
Multi-tenant support in Linear adapter

SamyPesse:
## Summary
The Linear adapter was simply not working in its current state: Not supporting multiple workspaces when using `clientId`/`clientSecret`; but more importantly throwing an error when trying to fetch the comments.

This PR makes the Linear adapter fully work, in multi-tenant and single-tenant modes.

- [x] Fix https://github.com/vercel/chat/issues/231: fix passing different scopes in single-tenant mode with `clientId`/`clientSecret` 
- [x] Fix https://github.com/vercel/chat/issues/345: fix handling installations in different workspaces when passing `clientId`/`clientSecret` (multi-tenant mode)
- [x] Fix `actor=app` mode where Linear will trigger an agent session (https://linear.app/developers/agent-interaction) instead of a normal comment
- [x] Support streaming, both for comments and agent sessions

## Test plan

Tests have been updated; but I also tested a build in my application.


dancer:
hi @SamyPesse looking into this & just saw a couple of things & ty for the PR!

getUserNameFromProfileUrl hardcoded to gitbook-x this needs updated or would break for others & startTyping missing return as in linear i believe they mention this somewhere on their docs user has 10 seconds to respond to a created even or the session is marked unresponsive 

also only one changeset file is needed major is fine!

SamyPesse:
> getUserNameFromProfileUrl hardcoded to gitbook-x this needs updated or would break for others

Oups, fixed.

> startTyping missing return as in linear

fixed.

> in linear i believe they mention this somewhere on their docs user has 10 seconds to respond to a created even or the session is marked unresponsive

Yes, I'm wondering if the adapter should not directly post an ephemeral message "Thinking" or if we wait for the first call to `startTyping` (what my app is doing).

dancer (approved):

PR #346, opened 2026-04-07
feat(chat): add thread.stream() method for passing streaming options

dancer:
## summary

`thread.stream()` was documented but didn't exist - users had to bypass the Thread abstraction and call `thread.adapter.stream()` directly to pass options like `taskDisplayMode` or `stopBlocks`

adds `thread.stream(stream, options)` which accepts `taskDisplayMode`, `stopBlocks`, and `updateIntervalMs` while auto-populating internal fields (`recipientUserId`, `recipientTeamId`) from message context

### DIFF
--- a/packages/chat/src/index.ts
+++ b/packages/chat/src/index.ts
@@ -335,6 +335,7 @@ export type {
   TaskUpdateChunk,
   Thread,
   ThreadInfo,
+  ThreadStreamOptions,
   ThreadSummary,
   WebhookOptions,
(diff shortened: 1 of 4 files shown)


cramforce:
Wow, really?

cramforce (changes requested):
Ah, now I get it. Not sure this should merge. We need to think about how to make this work across adapters.

We have the PostableObject abstraction for doing custom things

PR #348, opened 2026-04-08
feat: add unstorage state adapter

Schniz:
## Summary
- add a new `@chat-adapter/state-unstorage` package implementing the full `StateAdapter` surface on top of `unstorage`, with JSDoc for in-IDE docs
- include a comprehensive test suite mirroring state adapter compatibility scenarios (locks, cache, lists, queue, connection behavior) and switch subscriptions to per-key storage (`sub:<threadId>`) to avoid list RMW
- wire the new package into workspace/integration checks and update docs/catalog references across README, docs pages, adapters catalog, and chat skill table

## Validation
- `pnpm --filter @chat-adapter/state-unstorage test`
- `pnpm --filter @chat-adapter/state-unstorage build`
- `pnpm --filter @chat-adapter/integration-tests test`
- `pnpm validate`


Schniz:
this might not be good enough to ship because any read-manipulate-write code path will require distributed locking, which is not really possible with unstorage's api

at least until we will have un**lock** <picture data-single-emoji=":hehehe:" title=":hehehe:">:hehehe:</picture> 

could've been nice if we could ship that somehow though. maybe someone else has an idea?

PR #349, opened 2026-04-08
feat(chat): Enable disabling code-fencing for streamed tables per-adapter

raimondlume:
## Summary

As Slack now natively supports markdown tables, the code-fencing for streamed tables is not required and in fact stops them from being rendered natively.

This PR adds an option to configure code-fencing tables per adapter. By default `true`, `false` only for Slack

## Test plan

- Tests to validate `wrapTablesForAppend` behaviour
- Tables stream in natively in Slack instead of code fencing

Before:
image

After:
image

### DIFF
--- a/packages/chat/src/streaming-markdown.ts
+++ b/packages/chat/src/streaming-markdown.ts
@@ -1,5 +1,15 @@
 import remend from "remend";
 
+interface StreamingMarkdownRendererOptions {
+  /**
+   * Wrap confirmed table blocks in code fences for append-only consumers that
+   * cannot render markdown tables while a stream is in flight.
+   *
+   * Defaults to true to preserve existing behavior for current consumers.
+   */
+  wrapTablesForAppend?: boolean;
+}
+
 /**
  * A streaming markdown renderer that buffers potential table headers
  * until confirmed by a separator line, preventing tables from flashing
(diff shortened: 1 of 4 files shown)


packages/chat/src/streaming-markdown.ts
```diff
@@ -1,5 +1,15 @@
 import remend from "remend";
 
+interface StreamingMarkdownRendererOptions {
+  /**
+   * Wrap confirmed table blocks in code fences for append-only consumers that
+   * cannot render markdown tables while a stream is in flight.
+   *
+   * Defaults to true to preserve existing behavior for current consumers.
```
raimondlume: > Defaults to true to preserve existing behavior for current consumers.

remove

cramforce:
Last time I tested Slack did not support tables in streaming rendering

raimondlume:
> Last time I tested Slack did not support tables in streaming rendering

@cramforce I believe this was added [recently](https://docs.slack.dev/changelog/2026/03/06/block-kit-rich-text), see video below for example on how this looks in practice

https://github.com/user-attachments/assets/fdc4564c-2424-4713-9083-38f83964860c

cramforce:
Oh, very nice!

cramforce (approved):

cramforce:
@dancer Could you give this a quick try and merge?

PR #366, opened 2026-04-12
feat(telegram): expose parseMode config option

Koshkoshinsk:
## Problem

Since #245, the Telegram adapter hardcodes `parse_mode: "Markdown"` whenever the caller passes `{ markdown: ... }`. Telegram considers legacy `Markdown` a deprecated mode (see [Bot API formatting options](https://core.telegram.org/bots/api#formatting-options)) and it has no escape support, so callers can hit `Bad Request: can't parse entities` with no way to recover other than dropping the message permanently.

For example, an agent reply containing routine markdown like:

```
- **Trigger required?** (does every message need the trigger…)
- **Folder** name (e.g. `telegram_family-chat`)
```

is rejected by Telegram, and there is currently no way for the caller to pick a different parse mode or disable parsing entirely.

## Change

Adds an optional `parseMode` field to `TelegramAdapterConfig`:

```ts
parseMode?: "Markdown" | "MarkdownV2" | "HTML" | "none";
```

- Default remains `"Markdown"` — fully backward compatible with #245.
- `"MarkdownV2"` / `"HTML"` are passed through to Telegram unchanged.
- `"none"` omits `parse_mode`, sending content as plain text (useful as a safety valve).

The markdown converter is unchanged; this PR only exposes a knob. Callers whose upstream renderer produces MarkdownV2/HTML/plain-text can now tell the adapter not to ask Telegram to parse legacy Markdown entities.

## Tests

- New: `parseMode: "MarkdownV2"` and `"HTML"` round-trip into the request body.
- New: `parseMode: "none"` omits `parse_mode` entirely.
- New: invalid `parseMode` value throws `ValidationError` from the constructor.
- Existing parse_mode tests continue to pass with the unchanged default.

`pnpm turbo build --filter @chat-adapter/telegram` and `pnpm vitest run` (105 tests) both clean.

### DIFF
--- a/.changeset/telegram-parse-mode-config.md
+++ b/.changeset/telegram-parse-mode-config.md
@@ -1,0 +1,7 @@
+---
+"@chat-adapter/telegram": minor
+---
+
+Add `parseMode` config option to `TelegramAdapterConfig`. Allows callers to opt
+into `MarkdownV2`, `HTML`, or disable `parse_mode` entirely (`"none"`). Defaults
+to `"Markdown"` for backward compatibility.
--- a/packages/adapter-telegram/src/index.test.ts
+++ b/packages/adapter-telegram/src/index.test.ts
@@ -976,6 +976,90 @@ describe("TelegramAdapter", () => {
     expect(sendMessageBody.parse_mode).toBe("Markdown");
   });
 
+  it("honors parseMode config option for markdown messages", async () => {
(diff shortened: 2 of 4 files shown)


Koshkoshinsk:
Closing — going to handle this on our caller side instead, no upstream change needed for now.

Koshkoshinsk:
Reopening — apologies for the churn, the close was unintended on my end. Branch restored.

PR #367, opened 2026-04-12
feat(telegram): expose parseMode config option

Koshkoshinsk:
> Re-opens #366 — that PR was closed in error and GitHub will not let me reopen it after the branch was deleted. Same commit, same diff.

## Problem

Since #245, the Telegram adapter hardcodes `parse_mode: \"Markdown\"` whenever the caller passes `{ markdown: ... }`. Telegram considers legacy `Markdown` a deprecated mode (see [Bot API formatting options](https://core.telegram.org/bots/api#formatting-options)) and it has no escape support, so callers can hit `Bad Request: can't parse entities` with no way to recover other than dropping the message permanently.

For example, an agent reply containing routine markdown like:

\`\`\`
- **Trigger required?** (does every message need the trigger…)
- **Folder** name (e.g. \`telegram_family-chat\`)
\`\`\`

is rejected by Telegram, and there is currently no way for the caller to pick a different parse mode or disable parsing entirely.

## Change

Adds an optional \`parseMode\` field to \`TelegramAdapterConfig\`:

\`\`\`ts
parseMode?: \"Markdown\" | \"MarkdownV2\" | \"HTML\" | \"none\";
\`\`\`

- Default remains \`\"Markdown\"\` — fully backward compatible with #245.
- \`\"MarkdownV2\"\` / \`\"HTML\"\` are passed through to Telegram unchanged.
- \`\"none\"\` omits \`parse_mode\`, sending content as plain text (useful as a safety valve).

The markdown converter is unchanged; this PR only exposes a knob. Callers whose upstream renderer produces MarkdownV2/HTML/plain-text can now tell the adapter not to ask Telegram to parse legacy Markdown entities.

## Tests

- New: \`parseMode: \"MarkdownV2\"\` and \`\"HTML\"\` round-trip into the request body.
- New: \`parseMode: \"none\"\` omits \`parse_mode\` entirely.
- New: invalid \`parseMode\` value throws \`ValidationError\` from the constructor.
- Existing parse_mode tests continue to pass with the unchanged default.

\`pnpm turbo build --filter @chat-adapter/telegram\` and \`pnpm vitest run\` (105 tests) both clean.

### DIFF
--- a/.changeset/telegram-parse-mode-config.md
+++ b/.changeset/telegram-parse-mode-config.md
@@ -1,0 +1,7 @@
+---
+"@chat-adapter/telegram": minor
+---
+
+Add `parseMode` config option to `TelegramAdapterConfig`. Allows callers to opt
+into `MarkdownV2`, `HTML`, or disable `parse_mode` entirely (`"none"`). Defaults
+to `"Markdown"` for backward compatibility.
--- a/packages/adapter-telegram/src/index.test.ts
+++ b/packages/adapter-telegram/src/index.test.ts
@@ -976,6 +976,90 @@ describe("TelegramAdapter", () => {
     expect(sendMessageBody.parse_mode).toBe("Markdown");
   });
 
+  it("honors parseMode config option for markdown messages", async () => {
+    for (const mode of ["MarkdownV2", "HTML"] as const) {
+      mockFetch
+        .mockResolvedValueOnce(
+          telegramOk({
+            id: 999,
+            is_bot: true,
+            first_name: "Bot",
+            username: "mybot",
+          })
+        )
+        .mockResolvedValueOnce(telegramOk(sampleMessage()));
+
+      const adapter = createTelegramAdapter({
+        botToken: "token",
+        mode: "webhook",
+        logger: mockLogger,
+        userName: "mybot",
+        parseMode: mode,
+      });
+
+      await adapter.initialize(createMockChat());
+
+      await adapter.postMessage("telegram:123", {
+        markdown: "**bold**",
+      });
+
+      const sendMessageBody = JSON.parse(
+        String(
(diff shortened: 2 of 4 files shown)


Koshkoshinsk:
Good catch — pushed 048b951 adding `TelegramParseMode` to the type re-export block in `src/index.ts`. Build + 105 tests still clean.

bensabic:
Thanks for the PR and the clear writeup — the underlying problem you hit was real, but it's now been fixed at the root rather than exposed as a knob.

Since this was opened, the adapter migrated off legacy `parse_mode: "Markdown"` entirely (efae68f1 and follow-ups): rendered content is now emitted as properly-escaped **MarkdownV2** (`packages/adapter-telegram/src/markdown.ts`), and the adapter automatically retries as plain text when Telegram still rejects a body with `can't parse entities`. Your example message (`**Trigger required?**`, `` `telegram_family-chat` ``, etc.) renders correctly on `main` today.

That also makes each of the proposed option values moot:
- `"MarkdownV2"` — now the default behavior.
- `"Markdown"` (the proposed default) — would now misrender, since the renderer emits MarkdownV2 escapes that legacy mode displays literally.
- `"none"` — the safety valve is built in via the automatic plain-text fallback.
- `"HTML"` — the renderer never emits HTML, so this would always mismatch.

The one adjacent idea with residual value — attaching a `parse_mode` to `{ raw }` payloads for callers who pre-render their own Telegram markup — is something we'd rather design as a per-message option if/when there's concrete demand, so we're not going to land an adapter-wide config for it now.

Closing as superseded by the MarkdownV2 migration. If you're still seeing `can't parse entities` errors on the latest `@chat-adapter/telegram`, please open an issue with the failing payload — that would be a renderer bug we want to fix.

PR #371, opened 2026-04-13
feat: support govslack

gwkline:
## Summary
- allow the slack API URL to be configured in order to support usage of the chat SDK in a govslack environment

## Test plan
- we run a patched version of this package currently :)


cumason123:
We use the above for a patched version. That being said, streamed results in gov-slack become empty messages despite working in commercial (this may be a local implementation issue - warrants further investigation):

CleanShot 2026-04-13 at 09 33 26@2x

bensabic:
Thank you for this! It prompted us to add support to all adapters via PR #377

cumason123:
Thanks all!

cumason123:
Hey @bensabic, do you have any idea when a new pnpm package will be released with these changes? Do we need to trigger a changeset?

PR #395, opened 2026-04-17
fix(slack): enrich link previews with unfurl metadata from attachments

dancer:
## summary

- add `attachments` field to `SlackEvent` interface for legacy unfurl data
- enrich `LinkPreview` objects with title, description, imageUrl, and siteName from Slack's attachment metadata
- selectively handle `message_changed` events for unfurl updates instead of blanket-ignoring them
- links fetched via `conversations.replies` / `conversations.history` now include full OG metadata for `toAiMessages`

### DIFF
--- a/packages/adapter-slack/src/index.ts
+++ b/packages/adapter-slack/src/index.ts
@@ -102,6 +102,7 @@ function findNextMention(text: string): number {
  * The timestamp in the URL has no dot (e.g., p1234567890123456).
  * Supports optional query parameters (e.g., ?thread_ts=...&cid=...).
  */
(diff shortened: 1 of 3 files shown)


packages/adapter-slack/src/index.ts
```diff
@@ -1880,6 +1902,30 @@ export class SlackAdapter implements Adapter<SlackThreadId, unknown> {
     this.chat.processMessage(this, threadId, factory, options);
   }
 
+  private handleMessageChanged(
```
visyat, line 2066: This looks incomplete? We log the `message_changed` event but don't update the message?

visyat (approved):

PR #407, opened 2026-04-20
fix(telegram): MarkdownV2 rendering + telegram-chat reference example

### LINKED ISSUE #226: Telegram adapter: parse_mode not set for markdown messages

serejke:
## Summary

Two things bundled in one PR:

1. **Fix**: Telegram adapter was rendering messages in an incompatible dialect — standard markdown (`**bold**`) shipped with `parse_mode: "Markdown"` (legacy, single-asterisk) — so every LLM-generated message containing `.`, `!`, `(`, `)`, `-` got rejected with `can't parse entities`. Fixes #226.
2. **Example**: a new `examples/telegram-chat/` reference bot that exercises the adapter end-to-end: MarkdownV2 rendering, interactive cards, reactions, file uploads, streaming edits. Doubles as a manual regression harness for adapter changes.

### Root cause
The adapter's `fromAst()` delegated to the SDK's generic `stringifyMarkdown()`, which emits **standard markdown** (`**bold**`, no escaping). That output shipped with `parse_mode: "Markdown"` — Telegram's **legacy** parser that uses `*bold*` (single asterisk) and has no escape rules. Two incompatible dialects glued together.

### Changes to `packages/adapter-telegram`

- Switch `TELEGRAM_MARKDOWN_PARSE_MODE` to `"MarkdownV2"`.
- Replace `fromAst()` with a dedicated AST → MarkdownV2 renderer (`markdown.ts`):
  - Single `*bold*` / `_italic_` / `~strike~` markers.
  - Context-aware escaping: 20-char matrix for normal text; only `` ` `` and `\` inside code blocks; only `)` and `\` inside link URLs.
  - Headings rendered as bold (MarkdownV2 has no heading syntax).
  - Ordered/unordered lists with escaped dashes/periods.
  - Blockquotes with per-line `>` prefix.
  - Tables pre-empted and rendered as ASCII code blocks (existing behavior preserved).
  - Explicit handlers for `linkReference`, `imageReference`, `definition`, `html` so nothing is silently dropped.
- **Route card fallback text through `fromMarkdown`** (not raw escape), with `boldFormat: "**"` passed to `@chat-adapter/shared`'s `cardToFallbackText`. Default `boldFormat` is `"*"` (Slack mrkdwn) which, fed back through a markdown parser, becomes italic — not bold — on Telegram.
- **Fix `resolveParseMode`** so every message routed through the format converter (`{markdown}`, `{ast}`, cards, JSX) gets `parse_mode: "MarkdownV2"`. Previously only `{markdown}` and cards were covered, so `{ast}` messages shipped without parse_mode and rendered asterisks literally.
- Clarify inbound-vs-outbound: docstrings on `applyTelegramEntities` / `escapeMarkdownInEntity` note they're the *inbound* path (Telegram entities → standard markdown for `parseMarkdown`) and are distinct from the new *outbound* MarkdownV2 renderer.

### Tests (74 → 148)

- Full 20-char MarkdownV2 escape matrix.
- Context-escape tests: code block (only `` ` ``, `\`), link URL (only `)`, `\`).
- Nested formatting: bold-in-italic, code-in-link, list-with-bold.
- Edge cases: empty input, whitespace-only, raw HTML.
- **End-to-end LLM-output corpus test** with a MarkdownV2 validity invariant — strips entities and code blocks, asserts every remaining special char is escaped.
- Regression guards in `index.test.ts` for the AST / plain-string / raw `parse_mode` paths and for card-title MarkdownV2 bold rendering.

### Changeset
`patch` bump on `@chat-adapter/telegram`.

## The example (`12d5435`)

A polling-mode Telegram bot at `examples/telegram-chat/` that exercises the adapter end-to-end. One command to run (`pnpm --filter example-telegram-chat start`), no webhook, no public URL, no external API keys.

Menu structure — three categorized sub-menus, inline-keyboard navigation:

- **Text & Markdown** — 6 curated markdown demos + realistic LLM corpus + streaming edit loop.
- **Cards & Actions** — interactive approval card (edits in-place on press), **callback_data size probe** (see limitation note below), LinkButton.
- **Media & Reactions** — on-demand reaction one-shot (briefly subscribes), generated 1×1 PNG upload, generated minimal PDF upload.

Zero new runtime deps. PNG/PDF are hand-rolled in memory (`lib/png.ts` / `lib/pdf.ts`) rather than pulled from a binary-processing library.

Excluded from npm release via `.changeset/config.json`.

## Known limitation (not fixed in this PR)

The size-probe demo surfaces a DX gap worth flagging for a future PR:

- `@chat-adapter/shared`'s button encoder wraps every `Button.id` in a `chat:{\"a\":\"<id>\",\"v\":\"<value>\"}` JSON envelope before writing `callback_data`.
- That envelope eats ~13 bytes of Telegram's 64-byte limit, leaving ~51 bytes for id+value combined.
- This is documented nowhere; developers discover it by hitting `ValidationError: Callback payload too large for Telegram (max 64 bytes)` — which is exactly what the size-probe demo teaches.

Possible follow-ups in a separate PR: document the effective budget in the adapter README; consider a pluggable `CallbackEncoder` hook so apps with short action ids can opt into a leaner encoding (drop `chat:` prefix, single-string instead of JSON) when they don't need round-trip `value`. Happy to propose this separately if maintainers agree.

## Test plan

- [x] `pnpm --filter @chat-adapter/telegram test` — 148/148 pass
- [x] `pnpm --filter @chat-adapter/telegram typecheck` clean
- [x] `pnpm --filter example-telegram-chat typecheck` clean
- [x] `pnpm dlx ultracite check` on both packages clean
- [x] `pnpm knip` clean
- [x] Manual end-to-end against a real bot token: main menu renders, every category opens, every demo posts without `400 can't parse entities`, approval card edits in-place on button press, size-probe shows safe card + teaching error for oversize, streaming demo visibly edits a single message through 8 frames

Fixes #226


serejke:
image

serejke:
While working on this I noticed the truncator was producing bad MarkdownV2 — unescaped dots, orphan `\`, unclosed `*bold*` or `` `code` `` when the slice landed mid-entity. That's what the new `truncateForTelegram` handles now.

Also had a look at the other adapters while I was here — Discord has the same truncation bug, WhatsApp silently splits into multiple messages, Slack/GChat/Teams don't check at all. Filed #408 to clean that up separately.

dancer (approved):

dancer:
ty @serejke amazing work, will be in next release cc: @bensabic

JiehoonKwak:
when would be the next release? can't wait!

wong2:
Please release this fix!

PR #453, opened 2026-05-06
fix(docs): scroll code blocks in adapter READMEs

onmax:
in the top part is before, the bottom part is after.

We can now scroll on x and see full content

https://github.com/user-attachments/assets/da48337f-6113-4513-9a33-0e1e62f9c426


dancer (approved):

dancer:
ty @onmax

PR #454, opened 2026-05-06
feat(chat): add callbackUrl to buttons and modals

dancer:
## summary

add `callbackUrl` prop to `Button` and `Modal`. when a button is clicked or modal submitted, the SDK POSTs action data to the callback URL in addition to firing existing handlers. enables awaitable button/modal patterns with webhook-based workflow engines

- callback tokens stored server-side in state adapter (30-day TTL), only a short token goes in the platform payload
- original button value stored alongside the callback URL in state, keeping encoded values to 21 chars across all platforms
- works on Slack, Teams, Google Chat, WhatsApp, Telegram, and Discord
- Discord encodes value in `custom_id` (100 char limit, throws `ValidationError` if exceeded)
- Telegram's 64-byte `callback_data` limit accommodated by token-only encoding
- callback POST runs in parallel with handler execution, awaited at the end for serverless safety
- modal callback POST uses `waitUntil` to avoid blocking the submit response
- `callbackUrl` processed in `thread.post()`, `postEphemeral()`, `schedule()`, `edit()`, and all `channel` equivalents

## test plan

- callback-url unit tests: encode/decode, processCardCallbackUrls (flat, nested sections, immutability), resolveCallbackUrl (stored, legacy format, missing), postToCallbackUrl (success, non-2xx, network error)
- chat.test.ts integration tests: action handler decode + POST, no-value token, handlers fire alongside callback, expired/unknown token preserves raw value, modal submit POST
- channel.test.ts: callback URL encoding on channel.post
- discord cards.test.ts: encode/decode round-trip, validation throws on overflow, card rendering with values

### DIFF
--- a/.changeset/honest-trees-rush.md
+++ b/.changeset/honest-trees-rush.md
@@ -1,0 +1,7 @@
+---
+"chat": minor
+"@chat-adapter/discord": patch
+"@chat-adapter/teams": patch
+---
+
+add callbackUrl support to buttons and modals
--- a/.changeset/honest-trees-rush.md
+++ b/.changeset/honest-trees-rush.md
@@ -4,4 +4,6 @@
 "@chat-adapter/teams": patch
 ---
 
-add callbackUrl support to buttons and modals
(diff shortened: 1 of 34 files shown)


bensabic (approved):

visyat (changes requested):
Looks good for the most part. The workflow example you added to the example app is very interesting – may be worth adding to the "Durable Chat Agents" guide as well. I would really like replay tests, and workflow integration tests ideally!!

visyat (approved):

PR #492, opened 2026-05-12
feat(chat): add `chat/ai` subpath for AI SDK utilities

bensabic:
## Summary

Introduces a dedicated `chat/ai` subpath as the home for every Vercel AI SDK helper that ships with Chat SDK. Importing from this subpath keeps the optional `ai` and `zod` peer dependencies out of bundles that don't use them.

### What's new

- **`createChatTools`** — exposes Chat SDK operations as ready-to-use AI SDK tools so an agent can read, post, react, edit, delete, and manage thread subscriptions across every adapter the supplied `Chat` instance has registered.
  - Write operations require user approval by default (`requireApproval: true`); toggle globally or per-tool.
  - Three presets — `reader`, `messenger`, `moderator` — scope the toolset.
  - Individual tools can also be cherry-picked (`import { postMessage, addReaction } from "chat/ai"`).
- **`toAiMessages`** (and the `Ai*` / `ToAiMessagesOptions` types) now live alongside the tools at `chat/ai`. The previous `chat` re-exports continue to work, but are flagged `@deprecated` with an editor hint pointing to the new home — migration is a one-line import change.
- **Docs** — new `/docs/ai` section between Usage and Adapters in the sidebar:
  - `/docs/ai` — Overview
  - `/docs/ai/ai-sdk-tools` — `createChatTools` guide
  - `/docs/ai/to-ai-messages` — `toAiMessages` reference
  - `/docs/ai/types` — Reference for every type exported from `chat/ai`
- **Example app** — `examples/nextjs-chat` now demos the new surface via a "Run Agent Demo" button on the welcome card and a free-form `/agent <prompt>` slash command (streaming, with a placeholder so users get immediate feedback in channel contexts where Slack's typing-status API is a no-op).

### Future plans

`createChatTools` currently exposes the cross-adapter Chat SDK surface only. A natural follow-up is to also support **platform-specific tools** — e.g. expose Slack-only `pin`/`unpin`, Discord-only thread archiving, GitHub-only issue commenting, etc., so users can further extend what their agent can do without dropping back to raw adapter calls. The shape would likely be additional opt-in factories under `chat/ai` (or per-adapter subpaths like `@chat-adapter/slack/ai`) that return tools layered on top of the platform-specific adapter clients, while keeping the cross-platform `createChatTools` API as the lowest common denominator.

### Coverage

- `createChatTools` orchestrator: 100% statements / 94.7% branches.
- Every tool factory's `execute()` is exercised end-to-end (29 tests in `index.test.ts`).
- `toAiMessages` keeps its existing 35-test suite covering role mapping, attachment handling, links, transforms, and unsupported-attachment fallbacks.
- Tools folder overall: 99.0% statements / 86.1% branches / 97.4% functions / 98.9% lines.


apps/docs/content/docs/ai/index.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+```ts
+import { createChatTools, toAiMessages } from "chat/ai";
+```
+
+Add the optional peers if you don't already have them:
+
+```bash
+pnpm add ai zod
+```
```
HugoRCD: Shouldn't we use the package-install component to get all the pm?
```suggestion
```package-install
ai zod
```
```

File: apps/docs/content/docs/ai/ai-sdk-tools.mdx [hunk 2]
```diff
@@ -0,0 +1,227 @@
...
+
+```ts
+import { createChatTools } from "chat/ai";
+```
+
+`ai` and `zod` are optional peer dependencies — install them if you haven't already:
+
+```bash
+pnpm add ai zod
+```
```
HugoRCD: Same here we should use the package-install
```suggestion
```package-install
ai zod
```
```

File: apps/docs/content/docs/ai/index.mdx [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+  for await (const msg of thread.messages) {
+    recent.push(msg);
+    if (recent.length >= 20) break;
+  }
+
+  const history = await toAiMessages(recent);
+  const result = await agent.stream({ prompt: history });
+  await thread.post(result.fullStream);
+});
+```
```
HugoRCD: I think we can just do that, right?

```suggestion
bot.onSubscribedMessage(async (thread, message) => {
  const { messages } = await thread.adapter.fetchMessages(thread.id, { limit: 20 });
  const history = await toAiMessages(messages);
  const result = await agent.stream({ prompt: history });
  await thread.post(result.fullStream);
});
```
```

File: apps/docs/content/docs/ai/ai-sdk-tools.mdx [hunk 4]
```diff
@@ -0,0 +1,227 @@
...
+  adapters: { slack: createSlackAdapter() },
+  state: createMemoryState(),
+});
+
+const result = await generateText({
+  model: "anthropic/claude-sonnet-4.6",
+  tools: createChatTools({
+    chat,
+    preset: "messenger",
+    requireApproval: false,
```
HugoRCD: For someone new to the project who isn't familiar with all this, wouldn't it be a good idea to add a quick note as a warning to say that they're handing control over to the agent?

```suggestion
    requireApproval: false, // unattended script, no human-in-the-loop needed
```

examples/nextjs-chat/src/lib/bot.tsx
```diff
@@ -598,6 +598,115 @@ bot.onSlashCommand("/test-feedback", async (event) => {
...
+  const tools = createChatTools({
+    chat: bot,
+    preset: "messenger",
+    requireApproval: false,
+  });
+
+  await thread.startTyping("Running agent...");
+  try {
+    const result = await generateText({
+      model: "anthropic/claude-4.5-sonnet",
```
HugoRCD: The model on the gateway is called anthropic/claude-sonnet-4.5 but might be worth to upgrade to the 4.6:

```suggestion
      model: "anthropic/claude-sonnet-4.6",
```

btw the fake `anthropic/claude-4.5-sonnet` is used in many other places out of the scope of this PR

examples/nextjs-chat/src/lib/bot.tsx
```diff
@@ -598,6 +598,115 @@ bot.onSlashCommand("/test-feedback", async (event) => {
...
+  }
+
+  const tools = createChatTools({
+    chat: bot,
+    preset: "messenger",
+    requireApproval: false,
+  });
+
+  const toolAgent = new ToolLoopAgent({
+    model: "anthropic/claude-4.5-sonnet",
```
HugoRCD:
```suggestion
    model: "anthropic/claude-sonnet-4.6",
```

packages/chat/src/ai/index.ts
```diff
@@ -0,0 +1,308 @@
...
+      approval("postChannelMessage")
+    ),
+    sendDirectMessage: sendDirectMessage(chat, approval("sendDirectMessage")),
+    editMessage: editMessage(chat, approval("editMessage")),
+    deleteMessage: deleteMessage(chat, approval("deleteMessage")),
+    addReaction: addReaction(chat, approval("addReaction")),
+    removeReaction: removeReaction(chat, approval("removeReaction")),
+    subscribeThread: subscribeThread(chat, approval("subscribeThread")),
+    unsubscribeThread: unsubscribeThread(chat, approval("unsubscribeThread")),
+  };
```
HugoRCD: All 18 tools are instantiated upfront regardless of preset. Write tools (with their approval config) are constructed even for preset: "reader" and then immediately filtered out. Maybe we should do something like this to avoid instantiating everything?

```suggestion
  const factories = {
    fetchMessages:         () => fetchMessages(chat),
    fetchChannelMessages:  () => fetchChannelMessages(chat),
    fetchThread:           () => fetchThread(chat),
    listThreads:           () => listThreads(chat),
    getThreadParticipants: () => getThreadParticipants(chat),
    getChannelInfo:        () => getChannelInfo(chat),
    getUser:               () => getUser(chat),
    startTyping:           () => startTyping(chat),
    postMessage:           () => postMessage(chat, approval("postMessage")),
    postChannelMessage:    () => postChannelMessage(chat, approval("postChannelMessage")),
    sendDirectMessage:     () => sendDirectMessage(chat, approval("sendDirectMessage")),
    editMessage:           () => editMessage(chat, approval("editMessage")),
    deleteMessage:         () => deleteMessage(chat, approval("deleteMessage")),
    addReaction:           () => addReaction(chat, approval("addReaction")),
    removeReaction:        () => removeReaction(chat, approval("removeReaction")),
    subscribeThread:       () => subscribeThread(chat, approval("subscribeThread")),
    unsubscribeThread:     () => unsubscribeThread(chat, approval("unsubscribeThread")),
  };

  const allTools = Object.fromEntries(
    Object.entries(factories)
    .filter(([name]) => !allowed || allowed.has(name))
    .map(([name, build]) => [name, build()])
  );
```

HugoRCD (approved):

dancer (approved):

Issue #414, opened 2026-04-23
Add queue-debounce concurrency strategy for burst-collapse at idle


fuxingloh:
### Problem Statement

A very common user behaviour on chat platforms is sending a burst of messages in quick succession — typing a thought across three short lines, forwarding a batch of links, or firing off follow-up questions before the first has been read. Any assistant-style bot built on chat-sdk wants to treat this as **one logical turn**: read everything the user just sent, respond once.

### Proposed Solution

Introduce `'queue-debounce'` — like `'queue'`, but the FIRST message in an idle thread also waits `debounceMs` so a burst that arrives while the lock is free collapses into a single handler invocation. The handler receives the latest message with `context.skipped` containing the rest of the burst. Subsequent bursts arriving while the handler is running collapse via the same `queue` drain mechanism.

### Alternatives Considered

Neither existing concurrency strategy expresses this cleanly:

**`"queue"`** — msg 1 acquires the lock and dispatches its handler immediately. By the time msgs 2 and 3 land, the handler is already running; they enqueue and drain together after msg 1 finishes (`message = msg3`, `context.skipped = [msg2]`). The user gets **two handler invocations for one burst**: the bot replies to the first sentence, then a second reply for the rest. Reads as "bot answered before I finished typing."

**`"debounce"`** — collapses the burst at idle correctly but only the last message in the window dispatches.

### Use Case

```typescript
const chat = new Chat({
  adapters: {
    //...
  },
  state: opts.state,
  // `queue-debounce` first message on an idle thread waits `debounceMs` so a 
  // burst landing in that window collapses into one handler call with 
  // `context.skipped` populated. Subsequent bursts arriving mid-handler drain 
  // through the normal queue path — no orphan bug like pure `debounce`. Lets 
  // the agent respond to "three messages in 1s" as one turn instead of 1 | 2+3.
  // Cost: +1s baseline latency on every DM, including lone messages.
  // Acceptable for an assistant.
  concurrency: { strategy: "queue-debounce", debounceMs: 1000 },
  userName: "openxyz",
  logger: "info",
});
```

### Priority

Critical

### Contribution

- [x] I am willing to help implement this feature

bensabic:
This was implemented in #495 and released in `chat@4.29.0` as the `burst` concurrency strategy:

```typescript
concurrency: { strategy: "burst", debounceMs: 1000 }
```

[...]

PR #495, opened 2026-05-12
feat(chat): add queue-debounce concurrency strategy

dancer:
## summary

adds an opt-in `burst` concurrency strategy for #414

when a thread is idle, the first message waits for `debounceMs`, messages that arrive during that window are queued, and the handler runs once with the latest message plus earlier burst messages in `context.skipped`

after the handler finishes, messages that arrived while it was running are drained like `queue`, so the latest queued message is processed with earlier queued messages in `context.skipped`

keeps existing `drop`, `queue`, `debounce`, and `concurrent` behavior unchanged

updates docs to cover `burst`, explain when to choose it over `debounce`, and document the related `MessageContext` behavior

### DIFF
--- a/.changeset/soft-jars-cover.md
+++ b/.changeset/soft-jars-cover.md
@@ -1,0 +1,5 @@
+---
+"chat": minor
+---
+
+add queue-debounce concurrency strategy
--- a/.changeset/soft-jars-cover.md
+++ b/.changeset/soft-jars-cover.md
@@ -2,4 +2,4 @@
 "chat": minor
 ---
 
-add queue-debounce concurrency strategy
+add burst concurrency strategy
--- a/apps/docs/content/docs/concurrency.mdx
+++ b/apps/docs/content/docs/concurrency.mdx
@@ -1,6 +1,6 @@
 ---
 title: Overlapping Messages
-description: Control how overlapping messages on the same thread are handled — queue, debounce, drop, or process concurrently.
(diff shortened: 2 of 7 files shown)


.changeset/soft-jars-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"chat": minor
+---
+
+add queue-debounce concurrency strategy
```
visyat: Can we rename this to `burst` or something descriptive? `queue-debounce` is extremely confusing.

apps/docs/content/docs/concurrency.mdx
```diff
@@ -57,11 +57,47 @@ A done     → drain: [B, C, D] → handler(D, { skipped: [B, C] })
 D done     → queue empty → release lock
 ```
 
+### Queue debounce
+
+Waits once before the first handler on an idle thread, then drains the queued burst like `queue`. The latest message is dispatched, and earlier messages in the burst are available as `context.skipped`.
+
+Use this for assistant-style bots where users often send one logical turn as several short messages inside a small window and you want one response with full context. The tradeoff is that even a lone message waits for `debounceMs` before the handler runs.
```
visyat: Could perhaps document why someone would choose one approach over the other. It is not immediately intuitive why someone would choose this over `debounce` mode for example. 

Enables debouncing burst messages while still keeping all skipped messages for context. Also enables proper flushing of queued messages sent during handler execution.

visyat (changes requested):
Looks good aside from the mode naming!!

visyat (approved):

PR #538, opened 2026-05-19
feat(slack): add webhook primitives subpath

dancer:
## summary

adds a lightweight `@chat-adapter/slack/webhook` subpath for lower-level Slack webhook handling without pulling in the full Slack adapter runtime

this exposes primitives for verifying Slack requests, reading signed webhook bodies, parsing Events API callbacks, slash commands, and interactive payloads, and returning provider-native continuation data for message-like payloads

this gives framework and infrastructure integrations a small Slack-owned surface they can use directly while keeping `chat`, Slack SDK clients, adapter state, dedupe, locks, and subscriptions out of the primitive path


visyat (approved):
Minor NIT: Can we move `webhook-*` into `src/webhook/*` instead?

PR #554, opened 2026-05-25
feat(state-d1): add Cloudflare D1 state adapter

kauandotnet:
## Summary
- Adds **`@chat-adapter/state-d1`**, a SQLite-backed `StateAdapter` for Cloudflare Workers D1: persistent subscriptions, distributed locking, per-thread FIFO queues, ordered lists, and TTL cache.
- Behaviour mirrors `@chat-adapter/state-memory` (null-as-miss, `ttlMs: 0` = no expiry, token-owned locks, keep-newest trims, list TTL refresh) and the SQL structure of `@chat-adapter/state-pg`. Compound operations use `db.batch()` for atomicity; expired rows are cleaned up lazily on read/write.
- Registers the adapter in docs (`apps/docs/adapters.json`, `official/meta.json`, `official/d1.mdx`), `knip.json`, and the `integration-tests` README/doc import allow-lists. Adds a changeset.

## Test Plan
- **48 tests against a real D1** via `@cloudflare/vitest-pool-workers` (miniflare/workerd): `pnpm --filter @chat-adapter/state-d1 test`. Covers lock stealing/wrong-token/expiry, null-as-miss, `ttlMs:0`, FIFO + keep-newest trims, list TTL refresh, queue depth excluding expired, connection gating.
- `pnpm konsistent`, `pnpm knip`, `pnpm check`, `pnpm turbo typecheck`, and `pnpm build:validate` (docs build incl. `d1.mdx`) all pass.

## Notes for maintainers
- **vitest is pinned `^4.1` for this package only** — `@cloudflare/vitest-pool-workers@0.16.x` needs the `vitest/runtime` subpath added in 4.1 (the rest of the repo is on 4.0.18).
- **`test` omits `--coverage`** — `@vitest/coverage-v8` imports `node:inspector`, which workerd doesn't provide. Coverage isn't collected for this package.
- The package is intentionally **not** in the root `vitest.config.ts` `projects` array (that node-pool aggregate can't host a workers-pool project); it runs via `turbo test` like every other package.
- Introduces the repo's first Cloudflare/Workers footprint (`@cloudflare/workers-types`, `@cloudflare/vitest-pool-workers`, `wrangler` as devDeps; no runtime deps beyond `chat`).
- Commits are currently **unsigned** (no local signing key) — happy to re-sign before merge.


bensabic:
Thank you for this @kauandotnet! However, this would be a better fit as a vendor-official or community adapter.

PR #561, opened 2026-05-28
feat(chat): strict reviver that errors on missing _type

arjunkmrm:
## Summary

Adds an opt-in `strictReviver()` that throws when `JSON.parse(payload, strictReviver())` encounters a Thread- or Message-shaped object missing the `_type` discriminator.

The default `reviver` silently returns the literal in that case, so a downstream `thread.post(...)` call later throws `TypeError: thread.post is not a function` deep inside delivery code — far from the actual mistake.

We hit this in production after refactoring a delivery path that used the `JSON.stringify(...) + JSON.parse(..., bot.reviver())` pattern to construct Thread refs by hand. Missing `_type` meant the reviver returned a plain object, the type assertion lied, and `.post()` blew up on first delivery. This change makes that mistake fail loud at the parse boundary instead.

## Behaviour

- `reviver` — unchanged, backward compatible.
- `strictReviver()` — wraps `reviver`. Detects Thread-shaped values (`id` + `channelId` + `isDM`) and Message-shaped values (`id` + `threadId` + `text`/`parts`) without `_type` and throws a descriptive error.

## Test plan

- [ ] Existing reviver tests still pass (no change to `reviver`).
- [ ] New test: `strictReviver()` throws on `{ id, channelId, isDM }` without `_type`.
- [ ] New test: `strictReviver()` is a no-op for properly-typed payloads.

Happy to add the tests in this PR if the direction looks right — opening as a small surface-area proposal first.

### DIFF
--- a/packages/chat/src/index.ts
+++ b/packages/chat/src/index.ts
@@ -60,7 +60,7 @@ export type {
   PostableObjectContext,
 } from "./postable-object";
 export { isPostableObject } from "./postable-object";
-export { reviver } from "./reviver";
+export { reviver, strictReviver } from "./reviver";
 export { StreamingMarkdownRenderer } from "./streaming-markdown";
 export {
   StreamingPlan,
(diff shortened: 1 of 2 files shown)


arjunkmrm:
Closing — will route via fork-only PR for internal review first.

PR #573, opened 2026-06-01
fix(slack): pass token through native stream stop

### LINKED ISSUE #570: Slack adapter: streamer.stop() called without token causes not_authed when first append never sets it
### Bug Description

In `@chat-adapter/slack@4.29.0`, `SlackAdapter.stream()` passes the bot token only on the **first** `streamer.append()` call (gated by a local `first` flag) and never on `streamer.stop()`. When a streaming response reaches `streamer.stop()` without that first append having flushed, the underlying `ChatStreamer` in `@slack/web-api` has `this.token === undefined`. Its `stop()` then calls `chat.startStream` with `token: undefined`, and Slack returns `not_authed`. The assistant reply never lands in Slack.

The fix is simple: always pass `token` on every `streamer.append()` and on `streamer.stop()`. `ChatStreamer.append/stop` set `this.token = opts.token` only when truthy, so re-passing is idempotent.

### Steps to Reproduce

1. Run a Chat SDK app with the Slack adapter in webhook mode, with a valid bot token in `SLACK_BOT_TOKEN`.
2. In a handler, call `await thread.post(fromFullStream(streamText(...).fullStream))`.
3. Trigger a code path where no `streamer.append({ token })` ever executes before `streamer.stop()`. Confirmed triggers:
   - Structured chunks arrive first and the first attempt fails inside the `try { ... }` block — the catch sets `structuredChunksSupported = false` but throws before `first = false`. Subsequent structured chunks are no-ops, and `finalDelta` from `renderer.finish()` is empty.
   - Markdown-only response where `pushTextAndFlush` repeatedly produced `delta.length === 0` (renderer buffered without committing) and `renderer.finish()` produced no new committable text.
4. Watch Slack return `not_authed` from `streamer.stop()` → `chat.startStream`.

Independently verified by calling `POST https://slack.com/api/chat.startStream` with no `Authorization` header — same `not_authed` payload — while the same bot token succeeds on the same call when passed explicitly.

### Expected Behavior

`streamer.stop()` succeeds and the streamed assistant message is posted to the Slack thread.

### Actual Behavior

`streamer.stop()` throws:

```
{
  ok: false,
  error: 'not_authed',
  response_metadata: { acceptedScopes: ['chat:write'] }
}
```

Stack:

```
at b.platformErrorFromResult
at H.apiCall
at d.stop                    // ChatStreamer.stop
at a.stream                  // SlackAdapter.stream
at a.handleStream            // Thread.handleStream
```

### Code Sample

```typescript
// Suggested fix in SlackAdapter.stream()
// Drop the `first` flag and always pass `token`. ChatStreamer.append/stop
// internally do `if (opts.token) this.token = opts.token;` so re-passing
// is idempotent.

const flushMarkdownDelta = async (delta) => {
  if (delta.length === 0) return;
  await streamer.append({ markdown_text: delta, token });
};

const sendStructuredChunk = async (chunk) => {
  if (!structuredChunksSupported) return;
  const committable = renderer.getCommittableText();
  const delta = committable.slice(lastAppended.length);
  await flushMarkdownDelta(delta);
  lastAppended = committable;
  try {
    await streamer.append({ chunks: [chunk], token });
  } catch (error) {
    structuredChunksSupported = false;
    this.logger.warn(/* ... */);
  }
};

// ...

const result = await streamer.stop({
  token,
  ...(options?.stopBlocks ? { blocks: options.stopBlocks } : {}),
});
```

### Chat SDK Version

4.29.0 (@chat-adapter/slack); 4.29.0 (chat)

### Node.js Version

24

### Platform Adapter

_No response_

### Operating System

_No response_

### Additional Context

- Slack app has `chat:write`, `assistant:write`, `app_mentions:read`, `channels:history`, `groups:history`, `im:history`, `mpim:history`, `users:read.email`, `users:read`, etc. — `auth.test` returns `ok: true` with the same token.
- `@slack/web-api@7.16.0` (>= 7.14.0 as documented).
- Currently running as a pnpm patch in production at https://github.com/jindraregal/budeto-monorepo/blob/main/patches/%40chat-adapter__slack%404.29.0.patch (commit https://github.com/jindraregal/budeto-monorepo/commit/1cb6afc) — happy to open a PR against `vercel/chat` if useful.
- Supersedes #569 (filed without the bug-report template).

dancer:
## summary

fixes #570

fixes Slack native streaming when the stream reaches `stop()` before a token-bearing append has flushed

`SlackAdapter.stream()` now passes the resolved bot token on every `streamer.append()` and on `streamer.stop()`, matching Slack's auth requirements and avoiding `not_authed` responses from `chat.startStream` / `chat.stopStream`

adds regression coverage for buffered markdown streams and repeated structured chunk appends

### DIFF
--- a/packages/adapter-slack/src/index.test.ts
+++ b/packages/adapter-slack/src/index.test.ts
@@ -7583,7 +7583,7 @@ describe("stream with empty threadTs", () => {
     ).rejects.toThrow(ValidationError);
   });
 
-  it("passes token to stop when markdown remains buffered", async () => {
+  it("passes token on stream stop", async () => {
     const adapter = createSlackAdapter({
       botToken: "xoxb-test-token",
       signingSecret: "test-signing-secret",
--- a/packages/adapter-slack/src/index.test.ts
+++ b/packages/adapter-slack/src/index.test.ts
@@ -7582,6 +7582,86 @@ describe("stream with empty threadTs", () => {
       })
     ).rejects.toThrow(ValidationError);
   });
+
+  it("passes token to stop when markdown remains buffered", async () => {
+    const adapter = createSlackAdapter({
+      botToken: "xoxb-test-token",
+      signingSecret: "test-signing-secret",
+      logger: mockLogger,
+    });
+    const append = vi.fn().mockResolvedValue(null);
+    const stop = vi.fn().mockResolvedValue({
+      ok: true,
+      ts: "1234567890.111111",
+    });
+    const chatStream = vi.fn().mockReturnValue({ append, stop });
+    mockClientMethod(adapter, "chatStream", chatStream);
+
+    async function* shortStream() {
+      yield "hello";
+    }
+
+    await adapter.stream("slack:C123:1234567890.000000", shortStream(), {
+      recipientUserId: "U123",
+      recipientTeamId: "T123",
+    });
+
+    expect(stop).toHaveBeenCalledWith(
+      expect.objectContaining({
+        token: "xoxb-test-token",
+      })
+    );
+  });
+
+  it("passes token on every stream append", async () => {
+    const adapter = createSlackAdapter({
+      botToken: "xoxb-test-token",
+      signingSecret: "test-signing-secret",
+      logger: mockLogger,
+    });
+    const append = vi.fn().mockResolvedValue({ ok: true });
+    const stop = vi.fn().mockResolvedValue({
+      ok: true,
+      ts: "1234567890.111111",
+    });
+    mockClientMethod(
+      adapter,
+      "chatStream",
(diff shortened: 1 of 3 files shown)


packages/adapter-slack/src/index.test.ts
```diff
@@ -7582,6 +7582,86 @@ describe("stream with empty threadTs", () => {
       })
     ).rejects.toThrow(ValidationError);
   });
+
+  it("passes token to stop when markdown remains buffered", async () => {
```
gr2m: nit: maybe rename test? I think with a single "hello" chunk the markdown does flush?


```suggestion
  it("passes token on stream stop", async () => {
```

gr2m (approved):

PR #598, opened 2026-06-08
[chat] adding stable id to link button action handlers

visyat:
Enabling overriding the `link:<url>` action ID for `LinkButton` events.

### DIFF
--- a/.changeset/some-books-jog.md
+++ b/.changeset/some-books-jog.md
@@ -1,0 +1,7 @@
+---
+"@chat-adapter/slack": minor
+"chat": minor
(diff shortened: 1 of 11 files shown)


.changeset/some-books-jog.md
```diff
@@ -0,0 +1,7 @@
+---
+"@chat-adapter/slack": minor
+"chat": minor
+"docs": minor
```
bensabic: @visyat Nit: docs aren't needed in the changeset

bensabic (approved):

PR #602, opened 2026-06-09
feat(slack): expose webClientOptions to configure the underlying WebClient

aqt-dev:
## summary

adds `webClientOptions` to `SlackAdapterConfig` so users can configure the underlying Slack `WebClient` instances

the options apply to both the default client and per-token clients used for multi-workspace requests, including settings such as `retryConfig`, per-request `timeout`, custom headers, and `rejectRateLimitedCalls`

`slackApiUrl` is intentionally excluded from `webClientOptions` because the existing `apiUrl` option remains the single configuration path for overriding the Slack Web API base URL

custom headers are cloned for each client because the Slack SDK adds authorization to the provided headers object, preventing credentials from leaking between token-bound clients

## test plan

- verified options reach the default Slack client
- verified options reach clients created through `withBotToken()`
- verified custom headers remain isolated between token-bound clients
- verified the original headers object is not mutated
- verified `apiUrl` continues to configure the Slack Web API base URL
- ran the Slack adapter tests, workspace typecheck, formatting checks, and full build

### DIFF
--- a/.changeset/slack-web-client-options.md
+++ b/.changeset/slack-web-client-options.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": minor
+---
+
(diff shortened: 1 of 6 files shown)


aqt-dev:
Hi @dancer (cc @gr2m) — thanks for all the recent work on the Slack adapter, it's been great to build on.

I ran into a case where the adapter's `WebClient` is constructed without a `retryConfig` or `timeout`, so it inherits the SDK default (`tenRetriesInAboutThirtyMinutes`). Under sustained rate limiting a single `chat.update` can block for ~30 min, which stalls a streaming bot, and there's currently no way to tune it from outside the adapter.

This PR just adds an optional `webClientOptions` passthrough on `SlackAdapterConfig` (forwarded to both the default and per-token clients) so callers can set `retryConfig`/`timeout`. It's additive and non-breaking, default behavior is unchanged. Tests and a changeset are included.

No rush at all, but would love your eyes on it whenever you have a moment. Happy to adjust the approach if you'd prefer it shaped differently.

dancer (approved):

PR #612, opened 2026-06-12
fix(telegram): start typing before message processing

### LINKED ISSUE #611: Telegram typing indicator starts too late
In Telegram private chats, the native typing indicator can start too late.

The adapter supports typing via Telegram Bot API `sendChatAction({ action: "typing" })`, but in practice the first typing signal may only happen after Chat SDK processing reaches downstream handler code that calls `thread.startTyping()`.

That means a slow handler can leave the user with several seconds of no feedback, then the Telegram `...` indicator appears only briefly before the response starts.

Expected behavior: if a private Telegram message webhook is accepted for processing, the adapter should probably send `sendChatAction("typing")` as soon as it knows the thread id, before handing the message to downstream processing.

Relevant code:

- `handleIncomingMessageUpdate` already has the Telegram thread id, then parses/caches the message and calls `chat.processMessage(...)` without sending a typing action first: https://github.com/vercel/chat/blob/9c936f87960a968c9fa6070cd3188f68c989a7ac/packages/adapter-telegram/src/index.ts#L579-L596
- `startTyping()` already maps to Telegram Bot API `sendChatAction` with `action: "typing"`: https://github.com/vercel/chat/blob/9c936f87960a968c9fa6070cd3188f68c989a7ac/packages/adapter-telegram/src/index.ts#L1083-L1090
- `Chat.processMessage(...)` then enters the generic async message-processing path: https://github.com/vercel/chat/blob/9c936f87960a968c9fa6070cd3188f68c989a7ac/packages/chat/src/chat.ts#L917-L943

I am filing this to confirm whether maintainers consider that a bug, intended behavior, or something that should be behind an adapter option.

onmax:
## Summary

Send Telegram typing actions immediately for private incoming message and slash command updates before handing the message to Chat SDK processing. This lets Telegram clients show the native `...` indicator during early processing instead of waiting for downstream handler code to call `thread.startTyping()`.

The change is scoped to private, non-bot Telegram messages and adds regression coverage for normal messages and slash commands plus a patch changeset.

Closes #611

## Test plan

- [x] `pnpm validate`
- [x] `pnpm --filter @chat-adapter/telegram test`
- [x] `pnpm --filter @chat-adapter/telegram typecheck`
- [x] `pnpm --filter @chat-adapter/telegram build`

## Checklist

- [x] `pnpm validate` passes

### DIFF
--- a/.changeset/telegram-early-typing.md
+++ b/.changeset/telegram-early-typing.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/telegram": patch
+---
+
+Send Telegram typing actions immediately when private message updates arrive.
--- a/packages/adapter-telegram/src/index.test.ts
+++ b/packages/adapter-telegram/src/index.test.ts
@@ -345,6 +345,61 @@ describe("TelegramAdapter", () => {
     expect(threadId).toBe("telegram:-100123");
     expect(parsedMessage.text).toBe("hello @mybot");
     expect(parsedMessage.isMention).toBe(true);
+    expect(
+      mockFetch.mock.calls.some(([input]) =>
+        String(input).includes("/sendChatAction")
+      )
+    ).toBe(false);
+  });
+
+  it("starts typing before processing private message updates", async () => {
+    const events: string[] = [];
+    mockFetch
+      .mockResolvedValueOnce(
+        telegramOk({
+          id: 999,
+          is_bot: true,
+          first_name: "Bot",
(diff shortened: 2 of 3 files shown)


Jonathancombs782 (changes requested):
Sign and verify all commits as well as validate pnpm passes to finish whole checklist

dancer (approved):

PR #632, opened 2026-06-20
fix: let Plan tasks run in parallel without implicit auto-completion

### LINKED ISSUE #630: Plan API: addTask() auto-completes all previous in_progress tasks, preventing parallel task execution
### Bug Description

The `Plan.addTask()` method automatically marks all existing tasks with `in_progress` status as `complete` when adding a new task. This behavior prevents users from maintaining multiple tasks in a pending or in-progress state simultaneously, which is necessary for parallel task execution workflows.

**Root Cause:**
The `addTask()` method contains a loop that iterates through all existing tasks and sets their status to `complete` if they are `in_progress`:
```typescript
for (const task of this._model.tasks) {
  if (task.status === "in_progress") {
    task.status = "complete";
  }
}
```

**Suggested Fix:**
Remove or make optional the auto-completion behavior in `addTask()`. Consider adding an option parameter like `{ autoCompletePrevious: boolean }` to control this behavior, or remove it entirely and let users explicitly manage task status via `updateTask()`.

### Steps to Reproduce

```typescript
const _plan = new Plan({ initialMessage: "Step 1" });
await thread.post(_plan);

const step_1 = _plan.currentTask;
_plan.updateTask({ id: step_1?.id, status: "pending", });

const step_2 = await _plan.addTask({ title: "Step 2" });  // completes previous pending task as well
_plan.updateTask({ id: step_2?.id, status: "pending", });

const step_3 = await _plan.addTask({ title: "Step 3" });  // completes previous pending task as well
_plan.updateTask({ id: step_3?.id, status: "pending", });

_plan.complete({ completeMessage: "Done" });
```

### Expected Behavior

Multiple tasks should be able to exist in `pending` or `in_progress` state simultaneously. When `addTask()` is called, it should not automatically complete existing tasks unless explicitly requested by the user.

### Actual Behavior

When `addTask()` is called, all existing tasks with `in_progress` status are automatically set to `complete` status (lines 206-210 in `plan.ts`). This forces a sequential completion pattern even when the user wants parallel task execution.

### Code Sample

```typescript

```

### Chat SDK Version

4.30.0

### Node.js Version

24

### Platform Adapter

Slack

### Operating System

macOS

### Additional Context

_No response_

ASaiAnudeep:
## Summary

Plan’s task list API always marked existing in-progress steps as complete whenever a new step was added. That made sense for simple sequential bots, but it blocked parallel work — even though the docs already showed a parallel pattern and per-task updates by ID were added earlier.

This PR adds an optional flag on task creation so callers can keep multiple steps in progress at once, while leaving the old sequential behavior as the default.

**Opt-out flag, default on**. We considered removing auto-completion entirely. That would’ve been cleaner for parallel use but would’ve broken existing sequential bots that rely on implicit “move to next step” behavior. Defaulting to the current behavior keeps upgrades safe; parallel callers pass the flag off.

**No broader API redesign**. Task completion stays explicit via status updates and the existing “complete plan” flow. The change is scoped to when a new task is appended.

## Test plan

Unit tests

## Checklist

- [x] `pnpm validate` passes

closes #630

### DIFF
--- a/.changeset/plan-add-task-auto-complete-previous.md
+++ b/.changeset/plan-add-task-auto-complete-previous.md
@@ -1,0 +1,5 @@
+---
+"chat": minor
+---
+
+Add `autoCompletePrevious` option to `Plan.addTask()`. Defaults to `true` so sequential workflows keep auto-completing existing in-progress tasks; pass `false` to keep multiple tasks in progress for parallel execution.
--- a/apps/docs/content/docs/streaming.mdx
+++ b/apps/docs/content/docs/streaming.mdx
@@ -217,11 +217,14 @@ await plan.addTask({ title: "Summarize findings" });
 await plan.complete({ completeMessage: "Done!" });
 ```
 
-By default `updateTask()` mutates the most recent `in_progress` task. Pass `{ id }` to target a specific task — useful when work runs in parallel or out of order:
+By default `updateTask()` mutates the most recent `in_progress` task. For parallel work, pass `autoCompletePrevious: false` to `addTask()` so earlier tasks stay in progress; then pass `{ id }` to `updateTask()` to target a specific task when updates run out of order:
 
 ```typescript
 const fetchTask = await plan.addTask({ title: "Fetch data" });
-const transformTask = await plan.addTask({ title: "Transform" });
+const transformTask = await plan.addTask({
+  title: "Transform",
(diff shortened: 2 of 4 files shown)


ASaiAnudeep:
@cramforce can you help review this PR

dancer (approved):

PR #634, opened 2026-06-20
fix(chat): stop crashing thread().post(asyncIterable) on lightweight handles

### LINKED ISSUE #631: `chat.thread(threadId).post(asyncIterable)` crashes with `TypeError: Cannot read properties of undefined (reading 'userId')`
### Bug Description

`Chat#thread(threadId)` returns a lightweight `Thread` handle that is documented for posting messages from outside a webhook. Internally, it casts `{}` to `Message` and passes it as the current message:

```ts
// packages/chat/src/chat.ts:1863
return this.createThread(adapter, threadId, {} as Message, false);
```

That empty object flows into `ThreadImpl._currentMessage`. When the caller posts a streaming message (`thread.post(asyncIterable)`), `handleStream` enters the `if (this._currentMessage)` branch because `{}` is truthy, and immediately dereferences `author.userId`:

```ts
// packages/chat/src/thread.ts:611-619
if (this._currentMessage) {                                       // {} is truthy → enters
  options.recipientUserId = this._currentMessage.author.userId;   // {}.author === undefined → crash
  options.recipientTeamId = this.extractSlackRecipientTeamId(
    this._currentMessage.raw
  );
}
```

This throws `TypeError: Cannot read properties of undefined (reading 'userId')`.

The same `{} as Message` cast also exists in `Chat#openDM` (`packages/chat/src/chat.ts:1750`), so DM threads opened that way are affected as well.

Non-streaming `thread.post(stringOrPostable)` paths do not exercise the same branch and are unaffected. Mention-driven flows are also unaffected because the adapter passes a real `Message` into `createThread`.

### Steps to Reproduce

1. Create a `Chat` instance with a Slack adapter (any adapter that supports `stream` reproduces it).
2. Obtain a lightweight thread handle outside of a webhook:
   ```ts
   const thread = chat.thread("slack:C123ABC:1234567890.123456");
   ```
3. Post a streaming message:
   ```ts
   await thread.post(asyncIterableOfStringChunks);
   ```
4. See the crash:
   ```
   TypeError: Cannot read properties of undefined (reading 'userId')
       at ThreadImpl.handleStream (packages/chat/src/thread.ts:612)
   ```

### Expected Behavior

`chat.thread(threadId).post(asyncIterable)` should post the streaming message successfully. When there is no real current message context (i.e. the caller did not arrive via a webhook), `handleStream` should simply skip the `recipientUserId` / `recipientTeamId` derivation — which is exactly what the existing `if (this._currentMessage)` guard was clearly written to do.

### Actual Behavior

`handleStream` crashes with `TypeError: Cannot read properties of undefined (reading 'userId')` before any platform API call is made. The lightweight `Thread` is unusable for streaming posts.

### Code Sample

```typescript
import { Chat } from "chat";
import { createSlackAdapter } from "@chat-adapter/slack";

const chat = new Chat({
  adapters: [createSlackAdapter({ /* ... */ })],
});

async function* tokens() {
  yield "Hello ";
  yield "world";
}

const thread = chat.thread("slack:C123ABC:1234567890.123456");
await thread.post(tokens());
// → TypeError: Cannot read properties of undefined (reading 'userId')
//     at ThreadImpl.handleStream (packages/chat/src/thread.ts:612)
```

### Chat SDK Version

4.31.0

### Node.js Version

24.16.0

### Platform Adapter

Slack

### Operating System

macOS

### Additional Context

_No response_

tsushanth:
Closes #631.

`Chat#thread(threadId)` and `Chat#openDM(user)` returned a lightweight `Thread` handle by threading `{} as Message` into `ThreadImpl` as the `currentMessage`. The empty object is truthy, so the `if (this._currentMessage)` branch inside `handleStream` ran on every streaming `thread.post(asyncIterable)` and immediately dereferenced `this._currentMessage.author.userId`, throwing:

```
TypeError: Cannot read properties of undefined (reading 'userId')
```

`ThreadImpl`'s `currentMessage` field is already typed `Message | undefined` and `handleStream` already handles the absent case correctly — the constructor just lied about the type. Widen `createThread`'s `initialMessage` parameter to `Message | undefined` and pass `undefined` at the two lightweight-handle call sites (`Chat#openDM` line 1750, `Chat#thread` line 1863). Non-streaming `thread.post(stringOrPostable)` and webhook-driven paths used a real `Message` already and stay unchanged.

A regression test in `packages/chat/src/chat.test.ts` asserts `thread.post(asyncIterable)` no longer throws on a handle returned by `chat.thread(...)` — `await expect(...).resolves.not.toThrow()`.

### Validation

- `pnpm --filter chat test`: 1025 passed (24 files), including the new regression case.
- `pnpm --filter chat typecheck`: clean.
- `pnpm check` (ultracite / biome across the repo): 607 files, no diff.
- A changeset (`chat: patch`) is included.

### Signing note

The commit is signed with an SSH key I haven't yet registered as a signing key on github.com, so the "Verified" badge will be missing. Happy to register the key (or rotate to one that's already on file) once I have the `admin:ssh_signing_key` scope on this token — let me know if there's a preferred workflow.

### DIFF
--- a/.changeset/lightweight-thread-stream-crash.md
+++ b/.changeset/lightweight-thread-stream-crash.md
@@ -1,0 +1,5 @@
+---
+"chat": patch
+---
+
+Fix `TypeError: Cannot read properties of undefined (reading 'userId')` when calling `chat.thread(threadId).post(asyncIterable)` or posting a stream to a thread returned by `chat.openDM(user)`. Both lightweight-handle paths previously threaded `{} as Message` into the `ThreadImpl` constructor; the truthy empty object satisfied the `if (this._currentMessage)` branch in the streaming code path and crashed when it tried to read `this._currentMessage.author.userId`. Pass `undefined` instead so the no-current-message branch is taken — non-streaming `thread.post(stringOrPostable)` and webhook-driven paths are unaffected.
--- a/packages/chat/src/chat.test.ts
+++ b/packages/chat/src/chat.test.ts
@@ -1828,6 +1828,23 @@ describe("Chat", () => {
         'Adapter "unknown" not found'
       );
     });
+
+    // Regression for #631: a lightweight thread handle has no current
+    // message context, so streaming via `thread.post(asyncIterable)` used to
+    // crash on `this._currentMessage.author.userId` after the constructor
+    // cast `{} as Message` and the truthy empty object satisfied the
+    // `if (this._currentMessage)` branch in `handleStream`. Passing
+    // `undefined` instead of the cast keeps that branch unentered.
+    it("should not crash on stream() from a lightweight thread handle", async () => {
+      const thread = chat.thread("slack:C123:1234.5678");
+      await expect(
+        thread.post({
+          async *[Symbol.asyncIterator]() {
+            yield "Hi";
+          },
+        })
+      ).resolves.not.toThrow();
+    });
   });
 
   describe("getUser", () => {
--- a/packages/chat/src/chat.test.ts
+++ b/packages/chat/src/chat.test.ts
@@ -1652,6 +1652,17 @@ describe("Chat", () => {
         "Hello via DM!"
       );
     });
+
+    it("should not crash on stream() from a lightweight DM thread handle", async () => {
+      const thread = await chat.openDM("U123456");
+      await expect(
+        thread.post({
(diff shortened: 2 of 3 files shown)


dancer:
fixed in https://github.com/vercel/chat/pull/633

PR #639, opened 2026-06-23
fix(slack): recover expired native streams

### LINKED ISSUE #637: Slack streaming drops buffered text when the streamed message is closed mid-stream (message_not_in_streaming_state)
## Bug Description

`SlackAdapter.stream()` fails fatally and drops buffered text when a Slack streamed message is closed mid-stream.

Slack appears to finalize a streamed message (`chat.startStream`) after a bounded lifetime — we consistently observed appends failing roughly five minutes after the stream started. Once Slack has closed the message, the next `chat.appendStream` rejects with `message_not_in_streaming_state`. The adapter's `flushMarkdownDelta` (in `packages/adapter-slack/src/index.ts`) has no handling for this, so the rejection propagates out of `stream()` and every delta buffered since the last successful flush is lost.

This is routine for long-running agent turns, which interleave thinking and tool calls between bursts of answer text over many minutes — the single streamed message cannot outlive Slack's window, and the turn ends up truncated mid-word with the continuation missing.

## Steps to Reproduce

1. Start a stream into a Slack thread via the Slack adapter (`adapter.stream(threadId, iterable, { recipientUserId, recipientTeamId })`).
2. Have the async iterable yield text, then keep the stream open past Slack's streamed-message lifetime (about 5 minutes from the first append), then yield more text.
3. The append after the window rejects with `message_not_in_streaming_state`, `stream()` throws, and the in-flight text never reaches Slack.

## Expected Behavior

When Slack closes the streamed message mid-stream, the adapter should transparently continue on a fresh streamed message (so a long turn simply spans several messages) with no text dropped.

## Actual Behavior

`stream()` rejects with `message_not_in_streaming_state`. The caller's turn fails and the buffered tail (everything since the last successful flush) is lost. In the thread, the message is truncated mid-word and the continuation is missing.

Observed error:

```
An API error occurred: message_not_in_streaming_state
code: 'slack_webapi_platform_error'
data: { ok: false, error: 'message_not_in_streaming_state' }
  at ChatStreamer.flushBuffer
  at ChatStreamer.append
  at SlackAdapter.stream
```

## Code Sample

```typescript
// A stream that outlives Slack's streamed-message window loses its tail.
const streamer = client.chatStream({ channel, thread_ts, recipient_user_id, recipient_team_id });
await streamer.append({ markdown_text: "first chunk", token });
// ... stream stays open past Slack's streamed-message lifetime (~5 min) ...
await streamer.append({ markdown_text: "second chunk", token });
// -> rejects: message_not_in_streaming_state, and the buffered text is lost
```

## Chat SDK Version

`@chat-adapter/slack` 4.30.0 (the unguarded append is also present on `main` today: `packages/adapter-slack/src/index.ts`, `flushMarkdownDelta`).

## Node.js Version

25.x

## Platform Adapter

Slack

## Operating System

Linux

## Additional Context

The unguarded call is `flushMarkdownDelta` in `packages/adapter-slack/src/index.ts` (around L3996), which does `await streamer.append({ markdown_text: delta, token })` with no error handling.

Suggested fix: catch `message_not_in_streaming_state`, open a fresh `chatStream`, and resume from the last persisted point so a long turn spans several messages with nothing dropped. One subtlety worth flagging: because `ChatStreamer` batches (`buffer_size` defaults to 256), the resume point must be the last successful flush, not the last delta handed to the streamer — otherwise the prior buffered (unflushed) text is dropped. The adapter can distinguish the two via `append`'s return value (non-null means it flushed to Slack; null means it only buffered), which avoids both loss and duplication.

The underlying buffer also lives in `@slack/web-api`'s `ChatStreamer`, which could be hardened directly (it owns the unflushed buffer, so the fix there is smaller), but the adapter should be resilient to a mid-stream close regardless.

Happy to open a PR for the adapter-level fix if that is the preferred location.

tdietert:
## Summary

Fixes #637.

Slack can close a native streamed message server-side before `SlackAdapter.stream()` has finished consuming a long-running response. When Slack returns `message_not_in_streaming_state`, the adapter now stops using the expired native stream, keeps consuming the remaining text, and writes the final accumulated markdown back to the same Slack message via `chat.update` using the stream timestamp exposed by `@slack/web-api`.

This preserves buffered markdown that may not have reached Slack yet, including text still held by `ChatStreamer`'s internal buffer. The PR also bumps `@slack/web-api` to `^7.17.0` for the `ChatStreamer.ts` getter and adds regression coverage for both append-time and stop-time stream expiry.

## Test plan

- `pnpm --filter @chat-adapter/slack typecheck`
- `pnpm --filter @chat-adapter/slack exec vitest run src/index.test.ts -t "stream with empty threadTs"`
- `pnpm --filter @chat-adapter/slack test`
- `pnpm validate`

## Checklist

- [x] `pnpm validate` passes

### DIFF
--- a/.changeset/slack-stream-expiry.md
+++ b/.changeset/slack-stream-expiry.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": patch
+---
+
+Recover expired Slack native streams with `chat.update` so buffered markdown is not dropped when Slack returns `message_not_in_streaming_state`.
--- a/packages/adapter-slack/package.json
+++ b/packages/adapter-slack/package.json
@@ -45,7 +45,7 @@
   "dependencies": {
     "@chat-adapter/shared": "workspace:*",
     "@slack/socket-mode": "^2.0.5",
-    "@slack/web-api": "^7.14.0",
+    "@slack/web-api": "^7.17.0",
     "chat": "workspace:*"
   },
   "devDependencies": {
--- a/packages/adapter-slack/src/index.test.ts
+++ b/packages/adapter-slack/src/index.test.ts
@@ -8052,6 +8052,107 @@ describe("stream with empty threadTs", () => {
       expect.objectContaining({ token: "xoxb-test-token" })
     );
   });
+
+  it("falls back to chat.update when a stream append expires", async () => {
+    const adapter = createSlackAdapter({
+      botToken: "xoxb-test-token",
+      signingSecret: "test-signing-secret",
+      logger: mockLogger,
+    });
+    const expiredError = Object.assign(new Error("stream expired"), {
+      code: "slack_webapi_platform_error",
+      data: { error: "message_not_in_streaming_state", ok: false },
+    });
+    const append = vi
+      .fn()
+      .mockResolvedValueOnce({ ok: true, ts: "1234567890.111111" })
+      .mockRejectedValueOnce(expiredError);
+    const stop = vi.fn().mockResolvedValue({
+      ok: true,
+      ts: "1234567890.111111",
+    });
+    const update = vi.fn().mockResolvedValue({
+      ok: true,
+      ts: "1234567890.111111",
+    });
+    mockClientMethod(
+      adapter,
+      "chatStream",
+      vi.fn().mockReturnValue({ append, stop, ts: "1234567890.111111" })
+    );
+    mockClientMethod(adapter, "chat.update", update);
+
+    async function* expiringStream() {
+      yield "hello\n\n";
+      yield "world";
+    }
+
+    const result = await adapter.stream(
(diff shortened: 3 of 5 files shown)


tdietert:
Closing in favor of #671.

The single-`chat.update` recovery here is the right *safety net* but the wrong *primary* behavior for long-running streams — #671 tracks reactive **rollover** onto a fresh streamed message (continuing live streaming across a new message), with `chat.update` kept underneath only as the last-resort fallback.

Orphaning this PR per that direction. The groundwork carries over to the rollover implementation: expiry detection on the append/stop paths, the `ChatStreamer.ts` getter usage, the `@slack/web-api` `^7.17.0` bump, and the append-/stop-time expiry regression tests.

PR #672, opened 2026-07-02
Fix Slack native stream rollover

### LINKED ISSUE #671: Slack: auto-roll a native stream onto a fresh message when Slack expires it mid-stream (long-running, out-of-process producers)
### Problem Statement

`SlackAdapter.stream()` consumes an `AsyncIterable` and appends to a single Slack **native streamed message** (`chat.startStream` → `chat.appendStream` → `chat.stopStream`). That model assumes the producer feeds the stream more or less continuously and finishes well within Slack's streamed-message lifetime.

Our producer doesn't fit that assumption, and I suspect most agent integrations won't either:

- The thing producing the stream is an **agent running in a separate process**. It emits a **stream of events** over a long, bursty timeline — text deltas, tool-call activity, and task/todo updates, interleaved — which we forward into `adapter.stream()`. The agent is deliberately **decoupled** from the process doing the Slack streaming (it does not live in the same process as the text streamer).
- These agents are **long-running**: a single turn routinely spans many minutes as the agent thinks, calls tools, and emits answer text in bursts.

Slack finalizes a native streamed message after a bounded lifetime (we consistently observe appends starting to fail ~5 minutes after the stream opens). Once Slack has closed it, the next `chat.appendStream` / `chat.stopStream` rejects with `message_not_in_streaming_state` (`slack_webapi_platform_error`). `SlackAdapter.stream()` has no handling for this today, so the rejection propagates out of `stream()` and everything buffered since the last successful append is lost — the reply truncates mid-turn.

Crucially, for this use case it is **not enough to merely avoid dropping the tail**. A long turn legitimately outlives any single streamed message, so the adapter needs to **roll the connection**: when Slack expires the streamed message, open a *new* streamed message on the same thread and continue streaming into it, so one long turn transparently spans however many messages it needs.

(Same underlying expiry as the existing reports #607 and #637 — this issue is specifically about the *roll-onto-a-fresh-stream* behavior, not just recovering the lost text.)

### Proposed Solution

Make `SlackAdapter.stream()` **reactively roll onto a fresh native stream** when Slack expires the current one. High-level shape of the simplest correct fix:

1. Treat `message_not_in_streaming_state` as a **recoverable signal**, not a fatal error — detect it on both the append path and the final stop.
2. On that signal: capture the current streamed message's `ts`, open a new `chatStream` on the same thread, and **re-attempt only the pending delta** (the text accumulated since the last successful append). A cumulative-offset bookkeeping (the previous message holds `[0, lastAppended)`, the new one starts at `lastAppended`) guarantees the fresh message *continues* the answer rather than re-sending what the frozen message already holds — no duplication, no loss.
3. Continue streaming normally into the new message; **repeat the roll** if the turn outlives that message too.
4. **Bound** the number of rolls. If rollover can't make forward progress, fall back to a single `chat.update` with the accumulated text so the tail is never dropped even in the pathological case.

Reactive (roll on the actual error) rather than proactive (roll on a timer before the limit): Slack doesn't publish the streamed-message timeout, nor whether it's measured from stream start (absolute) or last append (inactivity), so a proactive timer can't be tuned reliably. Reacting to the error is axis-agnostic — it recovers whenever the window closes.

This needs the read-only `ChatStreamer.ts` getter to capture the streamed message id for the roll boundary (and the fallback), which landed in `@slack/web-api` **7.17.0** (`slackapi/node-slack-sdk#2611`, which also documents that a stream can expire server-side and that the SDK itself does no auto-recovery). The adapter currently allows `^7.14.0`, so this implies bumping that floor to `^7.17.0`.

Two continuity edges worth calling out — both solvable, both needed for the rendered output to stay correct across a message boundary:

- **Structured chunks (task/plan cards):** re-seed the latest snapshot onto the new message so the progress card survives the roll (the old card is frozen on the old message).
- **Open markdown constructs:** if the boundary lands inside a code fence or table, re-open it on the new message so the continuation renders as intended rather than as raw text.

### Alternatives Considered

- **Single `chat.update` fallback on expiry** (the direction in #637): catch the error and write the accumulated text into the same message once, then stop. This prevents data loss but does *not* preserve live incremental streaming past the window, and it collapses the remainder into one message (subject to per-message size limits). It's a good last-resort safety net — I'd keep it *underneath* rollover as the final guard — but not the right primary behavior for genuinely long streams.
- **Don't use native streaming at all** — buffer the whole turn and deliver via `chat.postMessage` + `chat.update`. This dodges expiry entirely but gives up native streaming's incremental UX and the Assistant task/plan cards, and reintroduces manual per-message size chunking. Effective, but it works *around* the SDK rather than fixing it.
- **Proactive timer-based rollover** before the limit: rejected because the timeout/axis is undocumented and can't be tuned (see above).

### Use Case

```typescript
import { Chat } from 'chat';
// A long-running agent in a *separate process* emits an event stream over many
// minutes: text deltas interleaved with tool-call activity and task updates.
async function* agentTurn(): AsyncIterable<string | StreamChunk> {
  for await (const event of subscribeToOutOfProcessAgent(sessionId)) {
    if (event.kind === 'text') yield event.text; // answer deltas
    else if (event.kind === 'task') yield event.chunk; // task/plan card update
    // ...tool activity, etc.
  }
}

// Today: if the turn runs past Slack's streamed-message lifetime (~5 min), the
// next append throws `message_not_in_streaming_state`, stream() rejects, and the
// buffered tail is lost.
//
// Desired: stream() transparently rolls onto a fresh streamed message and keeps
// going, so one long turn simply spans several messages with no gap and no loss.
await adapter.stream(threadId, agentTurn(), { recipientUserId, recipientTeamId });
```

### Priority

Important

### Contribution

- [x] I am willing to help implement this feature

### Additional Context

- Observed error on the append after the window closes:
  ```
  message_not_in_streaming_state
  code: 'slack_webapi_platform_error'
  data: { ok: false, error: 'message_not_in_streaming_state' }
  ```
- Chat SDK: `@chat-adapter/slack` 4.32.0; declared `@slack/web-api` range `^7.14.0`.
- Related issues: #607 (expiry during long gaps between chunks), #637 (buffered text dropped when the streamed message is closed mid-stream). This request differs in that it asks for **connection rollover** (continue live-streaming across a new message), not only recovery of the dropped tail.

tdietert:
Fixes #671

## What problems was I solving

Slack native streamed messages can stop accepting `chat.appendStream` / `chat.stopStream` calls while a long-running agent turn is still producing output. Slack reports that state as `message_not_in_streaming_state`; previously the adapter rejected and dropped later output.

This PR makes the Slack adapter treat that documented error as the recovery boundary: when the current native streamed message is no longer streamable, the adapter continues the turn in a fresh streamed message in the same thread and retries only the pending chunk.

## What user-facing changes did I ship

- [packages/adapter-slack/src/index.ts](https://github.com/vercel/chat/pull/672/files#diff-58a1ceb0817cf057cbb93be75aeea1543612f0deb4757c768bc9acecbc5636c5) — Slack native streams now recover reactively from `message_not_in_streaming_state` by starting a fresh stream and retrying only the pending markdown delta.
- [packages/adapter-slack/src/index.ts](https://github.com/vercel/chat/pull/672/files#diff-58a1ceb0817cf057cbb93be75aeea1543612f0deb4757c768bc9acecbc5636c5) — The adapter now calls `chat.startStream`, `chat.appendStream`, and `chat.stopStream` directly instead of going through `ChatStreamer`, so it owns stream lifecycle and confirmed-offset bookkeeping explicitly.
- [packages/adapter-slack/src/index.ts](https://github.com/vercel/chat/pull/672/files#diff-58a1ceb0817cf057cbb93be75aeea1543612f0deb4757c768bc9acecbc5636c5) — Latest structured task/plan chunks are re-seeded on rollover; fallback updates only the current message segment to avoid duplicating text already emitted in earlier rolled messages.
- [packages/adapter-slack/package.json](https://github.com/vercel/chat/pull/672/files#diff-47881123d78314450ff001368c44fbc372b2c39877816a231c632a236a58917d) — Keeps the Slack Web API floor at `^7.17.0` for the native stream chunk API surface used by the adapter.

## How I implemented it

- Removed the `ChatStreamer`/`buffer_size: 1` path and the open-code-fence-prefix workaround from the PR.
- Added direct native stream state tracking:
  - active Slack stream `ts`
  - last known message `ts`
  - current message start offset
  - last confirmed markdown offset
  - latest structured chunk snapshots
  - bounded rollover count
- On append expiry, the adapter starts a new stream in the same thread, optionally re-seeds latest structured chunks, and retries exactly the pending delta.
- On stop expiry, the adapter treats an already-flushed stream as complete; if final blocks or unconfirmed tail need recovery, it uses `chat.update` on the latest stream segment rather than duplicating the full cumulative answer.
- Updated adapter and integration tests to mock/assert direct Web API streaming methods instead of `chatStream()`.
- Added a patch changeset for `@chat-adapter/slack`.

### Manual testing

- [ ] Start a Slack native stream, let Slack report `message_not_in_streaming_state`, and confirm later deltas continue in a fresh streamed message.
- [ ] Confirm task/plan progress chunks still appear after a rollover.
- [ ] Confirm stop-time expiry does not truncate an already-flushed response.

### Automated tests

```bash
pnpm --filter @chat-adapter/slack... build
pnpm --filter @chat-adapter/slack... typecheck
pnpm --filter @chat-adapter/slack test
pnpm check
pnpm validate
```

All of the above passed locally after the refactor.

## Changelog

`@chat-adapter/slack` now recovers expired Slack native streams by reactively rolling long-running replies onto fresh streamed messages without dropping pending text.

### DIFF
--- a/.changeset/slack-stream-rollover.md
+++ b/.changeset/slack-stream-rollover.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": patch
+---
+
+Recover Slack native streams reactively by starting a fresh streamed message when Slack reports the current stream is no longer in streaming state.
--- a/packages/adapter-slack/package.json
+++ b/packages/adapter-slack/package.json
@@ -45,7 +45,7 @@
   "dependencies": {
     "@chat-adapter/shared": "workspace:*",
     "@slack/socket-mode": "^2.0.5",
-    "@slack/web-api": "^7.14.0",
+    "@slack/web-api": "^7.17.0",
     "chat": "workspace:*"
   },
   "devDependencies": {
--- a/packages/adapter-slack/src/index.test.ts
+++ b/packages/adapter-slack/src/index.test.ts
@@ -17,6 +17,16 @@ import { createSlackAdapter, SlackAdapter } from "./index";
 
 const FILE_ID_PATTERN = /^file-/;
 
+function createStreamExpiredError(): Error & {
+  code: string;
+  data: { error: string; ok: false };
+} {
+  return Object.assign(new Error("message_not_in_streaming_state"), {
+    code: "slack_webapi_platform_error",
+    data: { error: "message_not_in_streaming_state", ok: false as const },
+  });
+}
+
 // Mock @slack/socket-mode
 const mockSocketStart = vi.fn().mockResolvedValue({});
 const mockSocketDisconnect = vi.fn().mockResolvedValue(undefined);
@@ -3643,8 +3653,10 @@ interface MockableClient {
     postEphemeral: ReturnType<typeof vi.fn>;
     update: ReturnType<typeof vi.fn>;
     delete: ReturnType<typeof vi.fn>;
+    startStream: ReturnType<typeof vi.fn>;
+    appendStream: ReturnType<typeof vi.fn>;
+    stopStream: ReturnType<typeof vi.fn>;
   };
-  chatStream: ReturnType<typeof vi.fn>;
   conversations: {
     open: ReturnType<typeof vi.fn>;
     replies: ReturnType<typeof vi.fn>;
@@ -7918,6 +7930,37 @@ describe("routeSocketEvent with options", () => {
 // ============================================================================
 
 describe("stream with empty threadTs", () => {
+  function mockStreamMethods(
+    adapter: SlackAdapter,
+    options?: {
+      startStream?: ReturnType<typeof vi.fn>;
+      appendStream?: ReturnType<typeof vi.fn>;
+      stopStream?: ReturnType<typeof vi.fn>;
+      update?: ReturnType<typeof vi.fn>;
+    }
+  ) {
+    const startStream =
+      options?.startStream ??
+      vi.fn().mockResolvedValue({ ok: true, ts: "1234567890.111111" });
+    const appendStream =
+      options?.appendStream ??
+      vi.fn().mockResolvedValue({ ok: true, ts: "1234567890.111111" });
+    const stopStream =
+      options?.stopStream ??
+      vi.fn().mockResolvedValue({
+        ok: true,
+        ts: "1234567890.111111",
+      });
+    const update = options?.update ?? vi.fn();
+
+    mockClientMethod(adapter, "chat.startStream", startStream);
+    mockClientMethod(adapter, "chat.appendStream", appendStream);
+    mockClientMethod(adapter, "chat.stopStream", stopStream);
+    mockClientMethod(adapter, "chat.update", update);
+
+    return { appendStream, startStream, stopStream, update };
+  }
+
   it("delegates to fallback before consuming the stream", async () => {
     const adapter = createSlackAdapter({
       botToken: "xoxb-test-token",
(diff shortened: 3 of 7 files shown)


bensabic:
Thanks for this, Thomas! Most of this was addressed in #698 and #684.

PR #682, opened 2026-07-07
feat(x): add X (Twitter) adapter

dancer:
## summary

new `@chat-adapter/x` adapter for X (Twitter), built on the X API v2 and the X Activity API. write bot logic once and reply to mentions, hold DM conversations, post from the account, and like posts, like the other Chat SDK adapters

what it supports:
- reply to public mentions (`post.mention.create`) and top-level posts via `channel.post`
- send and receive direct messages (`dm.received` / `dm.sent`)
- edit and delete owned posts, delete own DM events
- likes as the only reaction (`emoji.heart` or `"like"`)
- buffered streaming: accumulates an LLM stream and posts once instead of post+edit churn on a public timeline
- OAuth 2.0 user context with managed token refresh (rotating refresh token persisted in the state adapter, optional AES-256-GCM encryption)
- webhook CRC and `x-twitter-webhooks-signature` verification

key design decisions:
- DMs are threaded by the other participant's user id (`x:dm:{userId}`) because X DM webhooks carry no conversation id, only participants
- OAuth 2.0 only at runtime: DM send and read are verified to work on OAuth 2.0 user tokens, so no OAuth 1.0a in the adapter (subscription and webhook setup is one-time and handled in the X developer console)
- parsers were written against real captured payloads: mentions use the v2 shape (author hydrated in `includes.users`), DMs use the legacy Account Activity shape (`direct_message_events`, `message_create.message_data`, a `users` map, and no conversation id)

also includes the `chat/adapters` catalog entry, docs page, CLI scaffold spec, and `sample-messages.md` with real captured payloads

<details><summary>usage</summary>

```typescript
import { Chat } from "chat";
import { createXAdapter } from "@chat-adapter/x";

const bot = new Chat({
  userName: "mybot",
  adapters: { x: createXAdapter() },
});

  await thread.post(`hi @${message.author.userName}!`);
});

bot.onDirectMessage(async (thread) => {
  await thread.post("hello from X");
});
```
</details>

## test plan

- adapter unit tests pass against the real captured payload shapes, with regression tests for author-from-`includes` (mentions) and the legacy `direct_message_events` shape (DMs)
- real captured `post.mention.create` and `dm.received` payloads verified end-to-end through `handleWebhook`: signature verification, routing, author resolution, and participant threading, plus bad-signature rejection returns 401
- every write and read path fired live against the X API through the adapter: top-level post, reply to a mention, like and unlike, edit, delete, DM send, DM read, DM delete
- OAuth 2.0 managed token refresh exercised live (access and refresh token rotation)


packages/adapter-x/src/index.ts
```diff
@@ -0,0 +1,1487 @@
...
+  protected buildDmMessage(
+    raw: Extract<XRawMessage, { kind: "dm" }>,
+    threadId: string
+  ): Message<XRawMessage> {
+    const { dmEvent, sender } = raw;
+    const text = dmEvent.text ?? "";
+    // Only adapter-created events are isMe (same contract as posts): a tracked
+    // dm.sent echo is filtered by core, avoiding reply loops, while other
+    // account-owned events still route by source.
+    const isMe = this.isSelf(dmEvent.id);
```
tobiaslins: **Blocking**: this only recognizes dm.sent echoes that were sent by this exact process. In serverless/multi-instance deployments, the webhook can arrive after a cold start or on another instance, so the bot's own DM is treated as user input and can trigger a reply loop. We should classify sender_id === botUserId as self, or persist sent IDs in shared state.

packages/adapter-x/src/index.ts
```diff
@@ -0,0 +1,1487 @@
...
+        this._botUserId = this._botUserId ?? me.data.id;
+        if (!this.hasExplicitUserName && me.data.username) {
+          this._userName = me.data.username;
+        }
+      }
+      this.logger.info("X adapter initialized", {
+        botUserId: this._botUserId,
+        userName: this._userName,
+      });
+    } catch (error) {
```
tobiaslins, line 177: **Blocking**: if X_USER_ID is omitted and `/2/users/me` fails, initialization continues without `botUserId`. That breaks dm.sent echo handling and participant threading because otherParticipant cannot tell bot-sent DMs from inbound ones. Since docs present `X_USER_ID` as optional, this should either fail init or make DM echo handling robust without optional identity.

packages/adapter-x/src/index.ts
```diff
@@ -0,0 +1,1487 @@
...
+): { dmEvent: XDmEvent; sender?: XUser } | null {
+  if (!isRecord(payload)) {
+    return null;
+  }
+  const events = payload.direct_message_events;
+  if (!Array.isArray(events)) {
+    return null;
+  }
+  const wire = (events as XDmWireEvent[]).find(
+    (event) => event?.message_create
```
tobiaslins: **Non-blocking/follow-up**: **direct_message_events** is an array, but this only extracts the first message_create event. If X batches multiple DM events in one delivery, the rest are silently dropped. Consider returning/routing all message_create events and adding a regression test for a multi-event payload.

visyat (approved):

PR #687, opened 2026-07-08
docs: use xai/grok-4.5 across docs, guides, and examples

dancer:
## summary

- swap chat model strings to `xai/grok-4.5` across the docs site, the shipped guides, and the example bots, so the docs lead with the latest model
- apps/docs: ai overview, ai-sdk-tools, and streaming pages, the two landing-page code samples, and the live chat demo route (`app/api/chat/route.ts`, previously `openai/gpt-4.1-mini`)
- packages/chat/resources/guides: the seven guides that use a chat model (slack connect, slack + ai sdk, liveblocks, vercel blob, github code review, daily digest, ai gateway)
- examples: nextjs-chat and nuxt-chat bots
- deliberately left non-chat model strings as they were, since grok-4.5 cannot fill those roles: the `openai/text-embedding-3-small` embedding model and the `openai/gpt-4o-mini` reranker
- in the ai gateway guide, grok-4.5 is now the primary model but the fallback list stays cross-provider (`anthropic/claude-opus-4.8`, `google/gemini-3.1-pro-preview`) so the failover example still demonstrates real cross-provider fallback
- also fixed a pre-existing prose/code mismatch in that guide, the fallback prose said `claude-opus-4.7` while the code listed `4.8`, now aligned to `4.8`


Jaaneek (approved):
looks perfect to me

bensabic (approved):

PR #714, opened 2026-07-16
fix(adapter-slack): establish per-team token context in Socket Mode (multi-workspace)

xuchen:
## Summary

In **multi-workspace mode** (an `installationProvider` is configured and there is no default `botToken`), the Slack adapter's **Socket Mode** dispatcher never resolves a per-team bot token. Every downstream Web API call made while handling a socket event then throws:

```
AuthenticationError: No bot token available. In multi-workspace mode, ensure the webhook is being processed.
```

This makes multi-workspace apps that use Socket Mode (no public ingress) unable to handle **any** inbound event — messages, mentions, slash commands, or interactive/button payloads all fail.

## Root cause

`getToken()` reads the token from the `requestContext` `AsyncLocalStorage`. The **webhook** path (`handleWebhook`) populates it per request: for `event_callback`, slash commands, and interactive payloads it calls `resolveTokenForTeam(installationId)` (which consults `installationProvider`) and runs the dispatch inside `this.requestContext.run({ ...ctx }, …)`.

`routeSocketEvent` (Socket Mode) does not — all three cases dispatch directly:

```ts
case "events_api":     this.processEventPayload(payload, options);            // no token ctx
case "slash_commands": wrapAsync(this.handleSlashCommand(params, options));   // no token ctx
case "interactive":    this.dispatchInteractivePayload(payload, options);     // no token ctx
```

So with no default token, `getToken()` has nothing to return and throws. (The error text — "ensure the webhook is being processed" — points right at this: only the webhook path sets the context.)

## Fix

Mirror the webhook token-context handling in `routeSocketEvent` for all three socket event types. When there is no default token, resolve the installation token via `resolveTokenForTeam` and run the dispatch inside `requestContext.run`. Team id comes from `body.team_id` (events_api / slash) or `payload.team.id` (interactive); enterprise installs use `enterprise_id` + `is_enterprise_install`.

Single-workspace (default token) behaviour is unchanged — the new path is guarded by `!this.defaultBotTokenProvider`, and everything falls through to the existing direct dispatch otherwise.

## Testing

- New tests under `socket mode - routeSocketEvent › multi-workspace token context (installationProvider)` assert each socket event type (`events_api`, `slash_commands`, `interactive`) resolves the per-team token via `installationProvider` before dispatch.
- `pnpm --filter @chat-adapter/slack typecheck` ✅
- `pnpm --filter @chat-adapter/slack test` ✅ (587 passed)
- `pnpm --filter @chat-adapter/slack build` ✅
- `biome check` ✅
- Changeset included (`@chat-adapter/slack` patch).

Happy to adjust naming/structure to fit your conventions.

### DIFF
--- a/.changeset/slack-socket-multiworkspace-token.md
+++ b/.changeset/slack-socket-multiworkspace-token.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": patch
+---
+
+Establish the per-team bot token context in Socket Mode for multi-workspace apps. `routeSocketEvent` now resolves the installation token via `installationProvider` and runs the dispatch inside `requestContext` for `events_api`, `slash_commands`, and `interactive` events — matching the webhook path. Previously, Socket Mode dispatched events with no token in multi-workspace mode (no `botToken`, only an `installationProvider`), so every downstream Web API call threw `AuthenticationError: No bot token available. In multi-workspace mode, ensure the webhook is being processed.`
--- a/packages/adapter-slack/src/index.test.ts
+++ b/packages/adapter-slack/src/index.test.ts
@@ -7971,6 +7971,115 @@ describe("socket mode - routeSocketEvent", () => {
     expect(chatInstance).toHaveDispatched("processAction");
   });
 
+  describe("multi-workspace token context (installationProvider)", () => {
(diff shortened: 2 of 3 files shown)


dancer:
resolved in https://github.com/vercel/chat/pull/724

PR #721, opened 2026-07-16
fix(discord): angle-wrapped markdown autolinks

### LINKED ISSUE #692: Discord: markdown converter does not handle 'preview hidden' links
### Bug Description

In Discord, you can wrap a link in angle brackets to suppress its preview embed, however the Chat SDK just ignores this.

### Steps to Reproduce

```ts
// with discord adapter
event.channel.post({ markdown: "<https://google.com>" })
```

### Expected Behavior

Keep the angle brackets, message content: `"<https://google.com>"`

### Actual Behavior

Actual message content: `"https://google.com"`

### Code Sample

```typescript

```

### Chat SDK Version

4.33.0

### Node.js Version

24.15.0

### Platform Adapter

Discord

### Operating System

Windows

### Additional Context

I will say, I probably could pass the string alone as the first parameter and it would skip the markdown pasrsing, however I was using streams and those are force passed with the markdown object (https://github.com/vercel/chat/blob/main/packages/chat/src/channel.ts#L295)

ishaksebsib:
Fixes #692

## Summary

Fixes Discord markdown rendering for angle-wrapped autolinks so link previews can be suppressed.

Previously:

```ts
 await thread.post({ markdown: "<https://google.com>" });
```

was sent to Discord as:

```txt
 https://google.com
```
image

Now it is preserved as:

```txt
 <https://google.com>
```
image

## Test plan
- pnpm validate
- Tested locally against a real Discord bot

## Checklist

- [x] `pnpm validate` passes
- [ ] Documentation updated (or N/A)

### DIFF
--- a/.changeset/discord-angle-autolinks.md
+++ b/.changeset/discord-angle-autolinks.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/discord": patch
+---
+
+Preserve Discord angle-wrapped markdown autolinks so link previews can be suppressed.
--- a/packages/adapter-discord/src/index.test.ts
+++ b/packages/adapter-discord/src/index.test.ts
@@ -1535,6 +1535,31 @@ describe("postMessage", () => {
     spy.mockRestore();
   });
 
+  it("preserves angle-wrapped autolinks in markdown message payloads", async () => {
+    const mockResponse = new Response(
+      JSON.stringify({
+        id: "msg-autolink",
+        channel_id: "channel456",
+        content: "<https://google.com>",
+        timestamp: "2021-01-01T00:00:00.000Z",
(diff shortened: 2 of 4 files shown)


DeanMauro:
I like the simplicity of this approach but in order to suppress previews on the `[text](<link>)` notation as well, changes are needed when parsing markdown to AST (at least I think that's the only way). Opened a fix in #726

ishaksebsib:
> I like the simplicity of this approach but in order to suppress previews on the `[text](<link>)` notation as well, changes are needed when parsing markdown to AST (at least I think that's the only way). Opened a fix in #726

This PR is only for the `<link>` case. I thought the `[text](link)` issue was already resolved based on your comment in #692, but I see it's not. In that case, I'll close this PR.

PR #723, opened 2026-07-17
fix(whatsapp): model forwarded and product-inquiry context variants in WhatsAppInboundMessage

ricardolx:
## Problem

`WhatsAppInboundMessage.context` is typed as:

```ts
/** Context for quoted replies */
context?: {
  from: string;
  id: string;
};
```

But Meta's Cloud API webhook sends **mutually exclusive context shapes** depending on message origin ([text messages webhook reference](https://developers.facebook.com/documentation/business-messaging/whatsapp/webhooks/reference/messages/text)):

- **Quoted replies** (and interactions with a business message): `from` + `id`
- **Forwarded messages**: only `forwarded` ("only included if forwarded 5 times or less") or `frequently_forwarded` ("only included if forwarded more than 5 times") — **no `id`, no `from`**
- **Catalog product inquiries**: `from` + `id` + `referred_product`

Because the type declares `id` required whenever `context` is present, downstream code like `message.raw.message.context?.id.trim()` type-checks cleanly and then throws a `TypeError` at runtime the first time a user forwards a message to the bot — `context` exists, `id` doesn't. We hit exactly this in production code building reply-to handling on top of the adapter.

## Change

- All `context` fields are optional, and the forwarded/product-inquiry fields (`forwarded`, `frequently_forwarded`, `referred_product`) are added, with doc comments noting which variant each field belongs to.
- Changeset included (`@chat-adapter/whatsapp`: patch).

The adapter itself never reads `context` at runtime, so this is a type-only change. Consumers who currently dereference `context.id` unguarded will get a compile error after upgrading — intentionally, since that code is a latent runtime crash on forwarded messages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/whatsapp-context-variants.md
+++ b/.changeset/whatsapp-context-variants.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/whatsapp": patch
+---
+
+Fix the `WhatsAppInboundMessage.context` type to model all documented webhook variants. The type previously declared `context?: { from: string; id: string }`, but Meta's Cloud API sends mutually exclusive context shapes: quoted replies carry `from`/`id`, forwarded messages carry only `forwarded` or `frequently_forwarded` (no `id`), and catalog product inquiries add `referred_product`. Code narrowed by the old type could dereference `context.id` and crash at runtime on forwarded messages. All context fields are now optional and the forwarded/product-inquiry fields are included.
--- a/.changeset/whatsapp-context-variants.md
+++ b/.changeset/whatsapp-context-variants.md
@@ -1,5 +1,5 @@
 ---
-"@chat-adapter/whatsapp": patch
+"@chat-adapter/whatsapp": minor
 ---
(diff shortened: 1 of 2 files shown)


ricardolx:
Reopening this — I closed it prematurely shortly after opening, which was a mistake on my part. The underlying issue is still present on `main`: `WhatsAppInboundMessage.context` still declares `from` and `id` as required, so consumers narrowing on that type can still dereference `context.id` and hit a runtime `TypeError` on forwarded messages.

Rebased onto current `main` and added the DCO sign-off. The diff is unchanged — still type-only, 2 files. The adapter never reads `context` at runtime (the sole reference in `packages/adapter-whatsapp/src` is a log string), so there's no behavioral change here.

Happy to adjust the shape if you'd prefer a discriminated union over all-optional fields.

dancer (approved):

PR #284, opened 2026-03-22
AI SDK Tool Approvals

bensabic:
**What**

- Adds `thread.requestApproval()` — posts a Card with Approve/Deny buttons, persists the pending approval in the StateAdapter, and returns a Promise that resolves when the user clicks
- Adds `thread.runAgent()` — runs an AI SDK agent in a loop, intercepts `tool-approval-request` content parts, posts approval cards, and resumes the agent with `tool-approval-response` messages in a tool-role message
- Supports `needsApproval` tools from AI SDK 6
- Exports card builders (`buildApprovalCard`, `buildResolvedCard`) and helpers (`isApprovalActionId`, `parseApprovalAction`) for custom usage

**Reliability**

- Singleton validated before any side effects — fails fast if `registerSingleton()` wasn't called
- Registry and state persisted before posting the card to prevent race conditions on fast clicks
- Registry and state cleaned up if posting the card fails
- Second state write uses remaining TTL to prevent expiry drift on slow posts
- Approval actions (`__approval:*`) intercepted before user `onAction` handlers
- Restart recovery via `bot.onApprovalResponse()` for approvals that arrive after a server restart

**Examples**

Standalone approval:
```ts
const { approved, user } = await thread.requestApproval({
  id: "transfer-123",
  title: "Transfer Money",
  fields: { Amount: "$500", To: "acct_123" },
});

if (approved) {
  await transferMoney(500, "acct_123");
}
```

With an AI SDK agent:
```ts
const result = await thread.runAgent(agent, {
  prompt: history,
  approvalCard: (toolCall) => ({
    title: `Confirm ${toolCall.toolName}`,
    fields: toolCall.input,
  }),
});
const text = await result.text;
await thread.post(text);
```

**Tests**

- 87 tests across `approval.test.ts` (58 unit) and `approval-integration.test.ts` (29 integration)
- Covers card builders, registry, state persistence, `requestApproval()`, `runAgent()`, and `handleApprovalAction()`

PR #728, opened 2026-07-21
feat(chat): durable human-in-the-loop approvals via chat/workflow

bensabic:
Adds a `chat/workflow` subpath export with `requestApproval()`. This is the DX from #284, rebuilt on Workflow SDK so the approval survives deploys, restarts, and arbitrarily long waits. No in-memory promises, no approvals registry, no restart-recovery machinery: the workflow suspends on a webhook and resumes when a button is clicked.

`requestApproval()` posts a card with Approve/Deny buttons whose `callbackUrl` targets a `createWebhook()` URL, suspends the workflow until a decision (or optional durable-sleep timeout), validates approvers, finalizes the card in place with the outcome (removing the buttons, leaving an audit trail), and returns the decision.

```typescript
import { requestApproval } from "chat/workflow";
import type { Thread } from "chat";
export async function deployApproval(opts: { thread: Thread; version: string }) {
  "use workflow";
  const { approved, user, timedOut } = await requestApproval(opts.thread, {
    title: `Deploy ${opts.version}?`,
    fields: { Version: opts.version },
    timeout: "24h",
    approvers: ["U_ALICE", "U_BOB"],
  });
  if (approved) {
    await deploy(opts.version);
  }
}
```

Starting it from a handler is one line. `Thread` instances serialize across the workflow boundary automatically via the existing `@workflow/serde` hooks on `ThreadImpl` (requires `chat.registerSingleton()`):

```typescript
import { start } from "workflow/api";
  await start(deployApproval, [{ thread, version: parseVersion(message.text) }]);
});
```

**Details**

- `workflow` is a new **optional** peer dependency (same pattern as `ai`); the subpath is the only code that imports it
- Unauthorized clicks (when `approvers` is set) and unrecognizable payloads post a notice / are ignored, and the workflow keeps waiting
- On timeout the card is finalized as timed out and the result has `timedOut: true`
- Card builders (`buildApprovalCard`, `buildResolvedCard`) are exported for custom flows
- Verified the published `dist` preserves the `"use step"` directives and down-levels `using` correctly, so the app-side Workflow SDK compiler handles the library code
- Docs page under Interactivity; changeset (`chat` minor); 8 unit tests mocking the `workflow` primitives

**Deliberate deviation from #284:** no `thread.requestApproval()` method. The function must suspend at workflow level, so hanging it off `ThreadImpl` would make `workflow` a hard dependency of core (or require prototype patching). The standalone `requestApproval(thread, options)` keeps the dependency optional.


dancer (approved):

packages/chat/package.json
```diff
@@ -60,16 +64,21 @@
     "tsup": "^8.3.5",
     "typescript": "^5.7.2",
     "vitest": "^4.0.18",
+    "workflow": "^4.5.0",
```
VaguelySerious: Workflow v5 is going GA soon, so we'd prefer if you use the `5.0.0-beta.x` releases instead, though not a big issue since it's a dev dep

VaguelySerious (commented):
Will get to a more thorough review in a few hours, left one comment in the meantime

VaguelySerious (approved):

PR #793, opened 2026-08-07
feat: support consumed streams without messages

### LINKED ISSUE #792: Adapter.stream cannot report a fully consumed StreamingPlan without creating a message
## Bug description

`Adapter.stream()` currently has only two outcomes:

- return a `RawMessage` after creating a platform message;
- return `null` to decline streaming and let core run its buffered fallback.

There is no valid outcome for an adapter that fully consumes a `StreamingPlan` but intentionally creates no platform message. This happens with structured-only streams when the adapter ignores unsupported structured chunks or handles them out of band.

## Steps to reproduce

1. Pass a `StreamingPlan` that emits only a structured chunk such as `task_update` to `thread.post()`.
2. Let the adapter consume the stream without creating a platform message.
3. Return `null` because the adapter has no `RawMessage` to return.
4. Core treats `null` as "not handled" and re-consumes the already exhausted stream through the buffered `post()` + `edit()` fallback.

Minimal shape:

```ts
async stream(_threadId, chunks) {
  for await (const chunk of chunks) {
    if (chunk.type === "markdown_text") {
      // normal platform streaming
    }
    // Unsupported structured chunks are intentionally ignored.
  }

  return null;
}
```

## Expected behavior

Adapters should have an explicit way to report that a `StreamingPlan` was fully consumed without creating a platform message. Core should then skip buffered fallback and avoid adding a nonexistent message to thread history.

`null` should retain its current meaning: the adapter declined before consuming the stream, so core may safely run fallback.

## Actual behavior

Returning `null` after consumption triggers fallback against an exhausted stream. Depending on the adapter, this can attempt to post blank content, produce a placeholder, fail delivery, or delay the final response.

## Environment

- `chat`: 4.36.0 (also observed with 4.35.0)
- Node.js: 22.18.0
- OS: Windows 11

I have a focused fix with regression tests, documentation, and a changeset ready to submit.

wwlsy-nql:
## Summary

- add `STREAM_HANDLED_NO_MESSAGE` for adapters that fully consume a `StreamingPlan` without creating a platform message
- keep `null` as pre-consumption fallback delegation and reject the new result for bare async iterables, preserving their `SentMessage` contract
- add regression coverage for structured-only streams, documentation, and a minor changeset

Closes #792.

## Root cause

`Adapter.stream()` could previously return only a platform message or `null`. Core always interprets `null` as "not handled" and runs buffered fallback. If an adapter has already consumed a structured-only stream, fallback receives an exhausted iterable and may attempt to send blank or placeholder content.

## Test plan

- `pnpm --filter chat exec vitest run src/thread.test.ts` — 154 tests passed
- `pnpm --filter chat test` — 26 files and 1115 tests passed
- `pnpm --filter chat typecheck` — passed
- targeted Biome check for changed TypeScript files — passed
- direct `pnpm exec tsup` from `packages/chat` — JavaScript and declarations built successfully
- verified all six remote file blobs match the locally tested files

`pnpm validate` could not complete in this Windows checkout because the repository-wide Ultracite step reports CRLF diagnostics across 646 pre-existing files. The changed TypeScript files pass the targeted formatter/linter check, and the committed documentation was normalized to LF before opening this PR.

## Checklist

- [ ] `pnpm validate` passes (blocked by repository-wide CRLF diagnostics in this Windows checkout; focused checks pass)


wwlsy-nql:
Closing this draft after a deeper contract review.

The symptom is real, but this change addresses it at the wrong abstraction layer. The original case comes from a Mastra `toolDisplay` callback using a `task_update` chunk as a boundary for an existing streaming session. Mastra currently opens a new `StreamingPlan` whenever a function-form tool display returns `{ kind: stream }`, even when there is no active text session. The custom adapter then intentionally renders no message for that boundary-only stream.

More importantly, the proposed API is not implementable as documented: `Adapter.stream(threadId, textStream, options)` is not told whether the iterable originated from a `StreamingPlan`, while `STREAM_HANDLED_NO_MESSAGE` is accepted only for `StreamingPlan` and rejected for a bare async iterable. An adapter therefore cannot safely decide whether returning the sentinel is valid for an otherwise identical stream.

The correct direction is to add an existing-session-only/boundary policy to Mastra's streaming driver, so a tool display event can affect an active session without opening an empty one. That avoids changing Chat SDK's global adapter return contract and avoids downstream shared state or fallback patches.

Thank you, and apologies for the noise. I am withdrawing this PR rather than stacking further changes onto the wrong API boundary.

PR #833, opened 2026-08-17
feat(telegram): implement native message replies

grootbro:
`Thread.reply()` throws `NotImplementedError` on Telegram: the adapter has no `reply` method, even though the Bot API threads an answer to its question with `reply_parameters`.

`postMessage` takes an optional reply target and passes it to every send path — text, rich messages, documents, attachments and both media group variants — and `reply()` delegates to it, the same shape the WhatsApp adapter uses for this contract. The target is decoded through the existing `decodeCompositeMessageId`, so a target from another chat is rejected exactly as an edit would be.

`allow_sending_without_reply` is set: a deleted target degrades to an unthreaded message instead of failing the send.

Three tests cover it: the reference lands on a reply, a plain `postMessage` stays unthreaded, and a target from another chat is refused.

### DIFF
--- a/.changeset/telegram-native-replies.md
+++ b/.changeset/telegram-native-replies.md
@@ -2,4 +2,4 @@
 "@chat-adapter/telegram": minor
 ---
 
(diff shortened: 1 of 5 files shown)


bensabic (approved):

grootbro:
@bensabic Thanks for reviewing and merging the Telegram reply support!

PR #834, opened 2026-08-17
feat(telegram): treat a reply to the bot as a mention

grootbro:
Based on #833.

In a group a bot only sees messages that address it, and people address a bot by replying to it as often as by typing its handle. The adapter reported `isMention` for the handle but not for the reply, so a bot went quiet the moment the conversation moved to replies.

`mentionOnReply` turns that on. **Off by default** — the flag changes which messages report `isMention`, and a bot that deliberately answers only explicit mentions should keep the stricter behaviour. It also reads `TELEGRAM_MENTION_ON_REPLY`, so a deployment can set it without code, and the key is declared in the adapters catalog.

The check runs before the empty-text guard, so a reply carrying only a photo or a document counts too.

### DIFF
--- a/.changeset/telegram-mention-on-reply.md
+++ b/.changeset/telegram-mention-on-reply.md
@@ -2,4 +2,4 @@
 "@chat-adapter/telegram": minor
 ---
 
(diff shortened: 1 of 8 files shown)


bensabic (approved):

grootbro:
@bensabic Thanks for the review and merge!

PR #835, opened 2026-08-17
feat(telegram): parse stickers and animations

grootbro:
Based on #834.

A sticker carries no text, so it reached the handler as an empty message and looked like a delivery that had lost its body. An animation — the MP4 Telegram sends for a GIF — was not declared on the message type and was dropped on the floor.

A sticker now reports the emoji it stands for as the message text, plus an image attachment typed by its real format: WebP for a still one, WebM for a video sticker, TGS for a Lottie one. An animation arrives as a video attachment alongside the other media types.

### DIFF
--- a/.changeset/telegram-sticker-animation.md
+++ b/.changeset/telegram-sticker-animation.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/telegram": minor
+---
+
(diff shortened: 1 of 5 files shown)


bensabic (approved):

grootbro:
@bensabic Thanks for reviewing and merging the sticker and animation support!

PR #836, opened 2026-08-17
feat(telegram): describe locations, contacts, polls and dice

grootbro:
Telegram sends several message kinds with neither text nor a file. They reached the handler as empty messages: the content was in the payload, but anything reading `text` saw nothing and could not tell an empty delivery from a shared location.

A location, venue, contact, poll, dice, game, invoice and story now each produce a short literal description, in the same place a sticker produces its emoji:

```
📍 55.75, 37.61
📍 Central Library, 12 Main St
👤 Ada Lovelace +15551234567
📊 Lunch or dinner?
🎲 4
🎮 Corsairs
🧾 Yearly plan — 49.99 USD
📖 Story
```

The wording stays minimal and the structured payload is untouched on the raw message, so a handler that wants the coordinates or the poll options still has them.

Two Bot API quirks shape the implementation:

- A venue message also carries a top-level `location` field for backward compatibility, so the venue check runs first. Otherwise every venue would render as bare coordinates.
- An invoice's `total_amount` is in the currency's smallest unit, and the exponent varies per currency ([currencies.json](https://core.telegram.org/bots/payments/currencies.json)): JPY and Telegram Stars count whole units, BHD, IQD and JOD use three decimals, everything else two.

`sample-messages.md` gains fixtures for the new kinds, including the venue with its co-set location and a contact without a `last_name`.

### DIFF
--- a/.changeset/telegram-non-file-content.md
+++ b/.changeset/telegram-non-file-content.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/telegram": minor
+---
+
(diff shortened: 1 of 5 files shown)


bensabic (approved):

grootbro:
@bensabic Thanks for reviewing and merging this Telegram batch! 🙌

PR #846, opened 2026-08-21
fix(slack): keep alert attachment content on normalized messages

### LINKED ISSUE #608: Slack: non-unfurl attachment content (title/text/fields) is dropped during message normalization
## Summary

Slack messages from integrations (Sentry, PagerDuty, GitHub, most alerting tools) carry their real payload in `attachments` — `title`, `text`, `fallback`, and the `fields` array — and/or blocks. The Slack adapter only reads `attachments` to build link previews (`from_url`/`original_url` unfurls). Attachments that aren't link unfurls are discarded: their content appears nowhere on the normalized `Message` (`msg.text` is just the top-level text; `msg.attachments` is for files).

## Impact

An alert that visibly renders a title, description, and key/value fields in Slack normalizes to a `Message` whose `text` is only the one-line summary. Everything consuming normalized messages inherits the blindness — `thread.messages`, `toAiMessages`, and the `chat/ai` tools (`fetchMessages` returns gutted alert messages to an agent). The only workaround is parsing `msg.raw` per platform, which defeats the normalization layer.

Related: #536 reports the same class of issue on Teams (file cards dropped during normalization).

## Suggested fix

Fold non-unfurl attachment content into the normalized message — append `fallback`/`title`/`text`/`fields` into `text`/`formatted`, or expose a dedicated field (e.g. `attachmentText`). The adapter already iterates the attachments array during normalization.

amitvijapur:
## Summary

Fixes #608.

Slack integrations such as Sentry, PagerDuty and GitHub carry their real payload in an attachment's `title`, `text` and `fields`. The adapter only read attachments to build link-unfurl previews, so that content reached nowhere on the normalized `Message`: `msg.text` held only the top-level one-line summary, and every consumer inherited the gap, including `thread.messages`, `toAiMessages` and the `chat/ai` `fetchMessages` tool.

Non-unfurl attachment content is now folded into the text before the AST is assembled, so both `formatted` and the derived plain text carry it. This follows the same approach as #817, which preserved pasted tables.

Three decisions worth calling out for review:

- Link unfurls stay excluded, on the same grounds their blocks already are via `isForeignAttachment`: the content is not the message author's.
- `fallback` is used only when the attachment has neither its own fields nor blocks. It is otherwise a plain-text stand-in that duplicates content rendered elsewhere, and including it unconditionally would inject strings like `[no preview available]` into messages that already carry table blocks.
- Mentions inside attachment content are resolved on the async path only, matching how table cells are already handled in `resolvedContent`.

`fields` was missing from the `SlackEvent["attachments"]` type and has been added.

## Test plan

Three tests added in `packages/adapter-slack/src/index.test.ts`, covering alert content (title, text and fields), the `fallback`-only case, and exclusion of unfurl attachments. The alert test asserts both the sync `parseMessage` and async `parseSlackMessage` paths, following the existing table-attachment test.

- `vitest run src/index.test.ts` in `packages/adapter-slack`: 419 passed.
- Reverting only the source change and keeping the new tests makes two of the three fail with the reported symptom (`expected 'New alert' to be 'New alert\n\nTypeError: cannot read p…'`), confirming they exercise the bug rather than the implementation.
- `tsc --noEmit` clean, `ultracite check` clean on the changed files.
- `pnpm validate` passes except `create-chat-sdk#test`, which fails identically on `main` at 294b595 in my environment with `Cannot find package 'chat/adapters'`. That looks like local workspace resolution rather than anything in this change.

## Checklist

- [x] `pnpm validate` passes
- [x] Documentation updated (or N/A — no documented behaviour changes)


songkeys:
Really need this!

bensabic (approved):

PR #862, opened 2026-08-25
feat(slack): add Agent Sessions lifecycle and native stop

bensabic:
Migrates Slack's `agent_view` integration to the Agent Sessions lifecycle while preserving the legacy `assistant_view` compatibility path.

- Adds `agents.sessions.setStatus` and `agents.sessions.rename` support for processing, active, suspended, and closed sessions.
- Handles `agent_session_stopped` without taking the message lock, clears Slack's processing state, and dispatches `onAgentSessionStopped`.
- Adds cross-process turn cancellation through the configured state adapter and exposes the active turn as `thread.signal`.
- Handles `agent_session_title_changed` and automatically titles new agent conversations from their root message, with a configurable resolver.
- Propagates `session_status` through native stream completion and supports suspended human-in-the-loop turns.
- Updates Slack manifests, examples, API docs, fixtures, and migration guidance for the February 2027 `assistant_view` retirement.

Configure the Agent messaging experience and optional title resolver:

```ts
const slack = createSlackAdapter({
  agentView: true,
  sessionTitle: ({ text }) => text.split("\n", 1)[0]?.slice(0, 80) ?? null,
});
```

Pass the thread signal into model generation so Slack's native stop button cancels upstream work as well as message delivery:

```ts
bot.onDirectMessage(async (thread, message) => {
  await thread.startTyping();

  const result = await agent.stream({
    prompt: message.text,
    abortSignal: thread.signal,
  });

  await thread.post(result.fullStream);
});
```

React to session lifecycle events:

```ts
bot.onAgentSessionStopped(async (event) => {
  await releaseExternalResources(event.threadId);
});

bot.onAgentSessionTitleChanged(async (event) => {
  await syncTitle(event.threadId, event.title);
});
```

Leave a stream suspended when the agent needs user input or approval:

```ts
await thread.post(
  new StreamingPlan(result.fullStream, {
    sessionStatus: "suspended",
  })
);
```


packages/adapter-slack/src/index.ts
```diff
@@ -5219,9 +5513,14 @@ export class SlackAdapter implements Adapter<SlackThreadId, unknown> {
     try {
       result = await streamer.stop({
         token,
+        ...(this.agentView
+          ? { session_status: options?.sessionStatus ?? "active" }
```
gr2m, line 5516: I don't think `session_status` is supported by Slack? I might be missing something but I think you'll have to set status with `setSessionStatus` explicitly: https://github.com/vercel/chat/pull/862/changes#diff-58a1ceb0817cf057cbb93be75aeea1543612f0deb4757c768bc9acecbc5636c5R3679

packages/adapter-slack/src/index.ts
```diff
@@ -615,6 +619,27 @@ interface SlackAssistantContextChangedEvent {
   type: "assistant_thread_context_changed";
 }
 
+interface SlackAgentSessionStoppedEvent {
+  channel: string;
+  event_ts: string;
+  message_ts: string;
```
gr2m: should this be `streaming_message_ts: string[]`?
https://docs.slack.dev/reference/events/agent_session_stopped/#usage-info

packages/adapter-slack/src/index.ts
```diff
@@ -615,6 +619,27 @@ interface SlackAssistantContextChangedEvent {
...
+  team_id: string;
```
gr2m: There is no `team_id` in the [agent_session_stopped event](https://docs.slack.dev/reference/events/agent_session_stopped/#usage-info)

packages/adapter-slack/src/index.ts
bensabic: This is part of a new API that Slack has rolled out, more info here: https://docs.slack.dev/ai/agent-sessions

gr2m: got it cheers 👍🏼

gr2m (approved):

PR #865, opened 2026-08-25
fix(adapters): guard attachment downloads across slack, discord, telegram, and whatsapp

bensabic:
Follows up on #850, #856, and #859 by adopting the shared guarded downloader (`downloadAttachment` in `@chat-adapter/shared`) in the remaining adapters that fetch attachment bytes from event-supplied URLs.

- Slack, Discord, and WhatsApp attachment downloads now refuse private and internal addresses (as URL literals, through DNS resolution, and after redirects), cap responses at 25 MB, and time out after 30 seconds.
- Slack sends the bot token only on hops to trusted Slack origins, so a redirect can never carry it to another host, and keeps the HTML-login-page detection. A protected `createFileTransport()` override routes downloads through a proxy.
- WhatsApp keeps its access token on Meta's media hosts, and the configured Graph origin via the hosts allowlist; `downloadMedia()` accepts a custom transport.
- Telegram keeps downloads on the Web Fetch API because a downstream Cloudflare Workers consumer depends on portability (#828), enforcing the same 25 MB cap and 30-second timeout with web streams.
- `downloadAttachment` now resolves `headers` per hop (function form decides what each redirect target receives), forwards the resolved headers to custom transports, and accepts an `onResponse` hook that can reject a final response before its body is read.
- Adds "Inbound attachments" docs sections for all four adapters.

### DIFF
--- a/packages/adapter-whatsapp/src/index.ts
+++ b/packages/adapter-whatsapp/src/index.ts
@@ -1,7 +1,9 @@
 import { createHmac, timingSafeEqual } from "node:crypto";
 import {
   AdapterError,
+  type AttachmentTransport,
   cardToFallbackText,
+  downloadAttachment,
   extractCard,
   extractFiles,
   extractPostableAttachments,
@@ -1150,7 +1152,10 @@ export class WhatsAppAdapter
    *
    * @see https://developers.facebook.com/docs/whatsapp/cloud-api/reference/media#download-media
    */
-  async downloadMedia(mediaId: string): Promise<Buffer> {
+  async downloadMedia(
+    mediaId: string,
+    transport?: AttachmentTransport
+  ): Promise<Buffer> {
     // Step 1: Get the media URL
     const metaResponse = await fetch(`${this.graphApiUrl}/${mediaId}`, {
       headers: { Authorization: `Bearer ${this.accessToken}` },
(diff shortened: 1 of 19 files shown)


packages/adapter-whatsapp/src/index.ts
```diff
@@ -1178,21 +1183,26 @@ export class WhatsAppAdapter
...
-    if (!dataResponse.ok) {
-      this.logger.error("Failed to download media", {
-        status: dataResponse.status,
-        mediaId,
+    // Step 2: Download the actual file. The hosts allowlist keeps the
+    // access token from following a redirect off Meta's media hosts.
+    try {
+      return await downloadAttachment(mediaInfo.url, {
+        adapter: "whatsapp",
+        headers: { authorization: `Bearer ${this.accessToken}` },
```
gr2m: The `hosts` allowlist matches subdomains and does not constrain ports, so this static header can still be forwarded when a configured Graph origin redirects to a child hostname or the same hostname on another port. Validate every hop with the existing exact-origin and Meta-host policy before returning the authorization header.

```suggestion
        headers: (target) => {
          if (!isWhatsAppMediaUrl(target.href, this.graphApiUrl)) {
            throw new NetworkError(
              "whatsapp",
              "Refusing to send the access token to an untrusted media URL"
            );
          }
          return { authorization: `Bearer ${this.accessToken}` };
        },
```

gr2m: I'd also add a regression tests

gr2m (approved):

PR #876, opened 2026-08-31
fix(adapters): validate external request targets

bensabic:
Adapters now reject untrusted destinations before sending credentials, message content, or attachment requests.

Teams Connector and Graph calls stay on known Microsoft hosts, Instagram downloads stay on trusted Meta hosts, and Slack response URLs are checked before use.

XChat now handles CRC challenges itself and rejects tokens that could be reused to forge webhook signatures.

### DIFF
--- a/.changeset/validate-external-targets.md
+++ b/.changeset/validate-external-targets.md
@@ -1,0 +1,8 @@
+---
+"@chat-adapter/instagram": patch
+"@chat-adapter/slack": patch
+"@chat-adapter/teams": patch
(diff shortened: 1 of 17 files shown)


gr2m (approved):
changes look good.

Consider hardening `sendSlackResponseUrl()` as well, it still POSTs its message payload to any caller-supplied URL. I didn't dig deeper into whether that's another security concern, maybe a note in the docs will suffice to warn users not to blindly pass through user-provided URLs to it?

PR #886, opened 2026-09-01
fix(slack): bound structured stream chunk fields

dcbuild3r:
## Summary

Slack limits structured task and plan fields to 256 characters. Truncate IDs, titles, and plan titles safely, and split oversized task details/output into stable continuation chunks so large tool results do not make native streaming fail. Preserve the prior part count on subsequent task updates so existing continuation cards can be updated consistently.

## Test plan

- bunx vitest run packages/adapter-slack/src/index.test.ts (444 passed)
- bunx tsc -p packages/adapter-slack/tsconfig.json --noEmit
- bunx biome check on changed files

## Checklist

- [x] All commits are signed off for the DCO (git commit -s)
- [ ] pnpm validate passes (targeted package validation run)
- [x] Changeset added
- [x] Documentation updated (N/A; automatic enforcement of Slack platform limits)

### DIFF
--- a/.changeset/tidy-tasks-slack.md
+++ b/.changeset/tidy-tasks-slack.md
@@ -1,0 +1,5 @@
+---
+"@chat-adapter/slack": patch
+---
+
+Split oversized Slack task updates and truncate structured stream fields to Slack's 256-character limit.
--- a/packages/adapter-slack/src/index.test.ts
+++ b/packages/adapter-slack/src/index.test.ts
@@ -10728,6 +10728,57 @@ describe("stream with empty threadTs", () => {
       expect.objectContaining({ token: "xoxb-test-token" })
     );
   });
+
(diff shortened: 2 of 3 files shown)


bensabic:
Thanks for this. Before merging I ran a probe against chat.appendStream with a live bot token to pin down the 256 rule from the docs. Slack accepted every case: title, details, and output at 300, 1,000, 4,000, 12,000, and 40,000 characters, ids at 300 and 1,000, plan titles at 1,000, 300-character titles under task_display_mode "plan", `pending` status, and thirty 300-character cards in one append. The only rejection was sending markdown_text and chunks in the same request, which the streamer never does.

Could you share the payload and the error code you hit? If Slack enforces a limit in some condition I missed, I'd like to target that exactly rather than truncate fields Slack renders fine today.

PR #906, opened 2026-09-05
feat(teams): render tables as the Adaptive Card 1.5 Table element

mmahmed:
## Summary

The Teams adapter rendered `Table` as a `Container` of `ColumnSet`s, one per row with every column `stretch`, which Adaptive Cards draws as stacked columns: no grid lines, no column sizing, no header semantics. Adaptive Cards 1.5, the version the adapter already declares, has a native [`Table`](https://adaptivecards.microsoft.com/?topic=Table) element that Teams renders as an actual table with grid lines, weighted column widths, per-column alignment and an accessible header row.

- **`@chat-adapter/teams`** — `convertTableToElement` now builds the 1.5 `Table` element from `@microsoft/teams.cards` (`Table`, `TableRow`, `TableCell`, `ColumnDefinition`) through the library's typed option objects. One column definition per column; a bold header row with `firstRowAsHeaders: true` only when `headers` is non-empty (`headers: []` yields a headerless table); one wrapped `TextBlock` per cell with the adapter's emoji conversion applied as before. Ragged rows are padded with empty cells, mirroring `tableElementToAscii`, and a table with no columns at all emits no element, matching its empty ASCII fallback. The dependency-free `@chat-adapter/teams/cards` subpath emits the same element, and a parity test renders one table through both paths. The declared card version stays `1.5`.
- **`chat`** — `Table` / `TableOptions` / `TableProps` gain optional, platform-neutral rendering options, each documented "Rendered by Teams only; other adapters ignore it": `widths` (positive integer column weights, parallel to `align`; anything else falls back to 1, because Teams desktop and mobile disagree on zero, negative and fractional weights), `verticalAlign` (`"top" | "center" | "bottom"`), `gridLines` (default `true`) and `gridStyle` (`"default" | "emphasis" | "accent" | "good" | "attention" | "warning"`). They survive the factory, the chat JSX runtime and `fromReactElement`. Every field is optional, so no existing call site changes. `TableVerticalAlignment` and `TableGridStyle` are exported next to `TableAlignment`.
- The chat JSX runtime dropped `align` on `<Table>` because `TableProps` and the resolver kept their own copy of the option list. `TableProps` is now an alias of `TableOptions`, and both the JSX resolver and `fromReactElement` hand the props to `Table()`, which copies each option by name, so a new option is added in one place.
- Docs: the Table section and `TableOptions` reference, the Teams adapter page and its README feature row. Changeset bumps `chat` and `@chat-adapter/teams` as `minor`.

No other adapter reads the new fields; they consume `headers`, `rows`, `align`, `caption` and `pageSize` only, and none of their tests changed.

### Rendering change for every Teams table

Every existing `Table` on Teams now renders with grid lines (`showGridLines: true`): a table with lines is what the element means and what Teams draws for a markdown table. If you would rather existing cards keep a borderless look, I am happy to flip the default to `gridLines: false`; say so and I will update the default, tests and docs.

### `firstRowAsHeaders`

The published Adaptive Cards schema spells the header flag `firstRowAsHeader`, but the Teams renderers and `@microsoft/teams.cards` read the plural `firstRowAsHeaders` (default `true`) and ignore the singular. Microsoft confirmed in [msteams-docs#9241](https://github.com/MicrosoftDocs/msteams-docs/issues/9241) that `firstRowAsHeaders: false` renders a headerless table on Teams desktop, web and mobile; the schema discrepancy is tracked in [AdaptiveCards#8599](https://github.com/microsoft/AdaptiveCards/issues/8599). The adapter sets the flag through the library's typed options, and a comment at each emit site records the trap. Because the library defaults it on, a headerless table sets it to `false` explicitly.

## Test plan

- `pnpm check`, `pnpm typecheck`, `pnpm knip`, `pnpm konsistent`: clean.
- `pnpm --filter @chat-adapter/teams test`: new `cardToAdaptiveCard with Table` suite covering the native `Table` with grid lines on by default; default weight 1, `widths` on the column definitions, and the fallback to 1 for zero, negative, fractional and NaN weights; per-column `align` and table-level `verticalAlign` for every value; header row and `firstRowAsHeaders` only when headers exist; no element for a table with no columns; `gridLines: false`; `gridStyle`; ragged-row padding; emoji placeholders in cells; the ASCII fallback unchanged; and a parity test asserting the full adapter and the `@chat-adapter/teams/cards` subpath emit the same JSON. The `cards-primitives` suite covers the subpath the same way.
- `pnpm --filter chat test`: each new option survives `Table()`, the chat JSX runtime (`<Table>`) and `fromReactElement`.
- Full `pnpm test` across every package.
- `pnpm validate`.

## Checklist

- [x] `pnpm validate` passes

### DIFF
--- a/.changeset/teams-native-table.md
+++ b/.changeset/teams-native-table.md
@@ -1,0 +1,12 @@
+---
+"chat": minor
+"@chat-adapter/teams": minor
+---
+
+feat(teams): render tables as the Adaptive Card 1.5 Table element
+
+The Teams adapter now renders `Table` as the native Adaptive Cards `Table` element instead of a `Container` of `ColumnSet`s. Teams draws grid lines between cells, sizes columns by relative weight and marks the header row for accessibility. The `@chat-adapter/teams/cards` subpath emits the same element.
+
+`Table` gains optional rendering options that only Teams reads: `widths` (positive integer column weights), `verticalAlign` (vertical alignment of cell content), `gridLines` (default `true`) and `gridStyle`. Other adapters ignore them. Pass `gridLines: false` to keep a borderless table.
+
+The chat JSX runtime now forwards `align` on `<Table>`, matching `fromReactElement`.
--- a/apps/docs/content/adapters/official/teams.mdx
+++ b/apps/docs/content/adapters/official/teams.mdx
@@ -24,7 +24,7 @@ features:
   selectMenus: no
   tables:
     status: yes
-    label: GFM
+    label: Adaptive Card Table
   fields: yes
   imagesInCards: yes
   modals: yes
@@ -522,6 +522,8 @@ await postTeamsMessage({
 ```
 
 Use the full Chat SDK card JSX when you want cross-platform rendering. Use `@chat-adapter/teams/cards` when you are building a Teams-only runtime and want Adaptive Card output directly.
+
+Table children render as the Adaptive Card `Table` element with grid lines on by default. The optional `align`, `widths`, `verticalAlign`, `gridLines` and `gridStyle` fields tune the grid, a table with empty `headers` renders without a header row, and a table with no columns emits no element.
 
 ### Modals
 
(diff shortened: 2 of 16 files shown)


mmahmed:
@bensabic - please review and let me know if you need any changes here

bensabic (approved):

PR #910, opened 2026-09-06
fix(whatsapp): stop escaping literal tildes and rewriting markers in code content

### LINKED ISSUE #908: WhatsApp adapter adds visible backslashes before literal tildes
### Bug Description

In `@chat-adapter/whatsapp@4.39.0`, outgoing Markdown containing a literal tilde gets an extra backslash. For example, `(~80 % easy)` is delivered as `(\~80 % easy)`, with the backslash visible in WhatsApp.

The original model output contains no backslash. This reproduces directly in the adapter’s formatter without an LLM or network request.

### Steps to Reproduce

1. Use `@chat-adapter/whatsapp@4.39.0` with `chat@4.37.0` (our project pins `chat` via an override).
2. Run the minimal code sample below with Bun.
3. Observe that the formatter returns `(\~80 % easy)` instead of `(~80 % easy)`.
4. Sending `{ markdown: "(~80 % easy)" }` uses the same converter and includes the unwanted backslash in the outgoing message.

### Expected Behavior

The formatter should return `(~80 % easy)` unchanged. Literal text should arrive without serializer-added Markdown escapes, while supported formatting still converts to WhatsApp syntax. Intentional backslashes and code content should remain intact.

### Actual Behavior

The formatter returns `(\~80 % easy)`. The added backslash remains visible in the delivered WhatsApp message.

This also affects multiple tildes: `~45 min, Pace ~6:29/km` becomes `\~45 min, Pace \~6:29/km`.

### Code Sample

```typescript
import { WhatsAppAdapter } from "@chat-adapter/whatsapp";
import { ConsoleLogger, parseMarkdown } from "chat";

const adapter = new WhatsAppAdapter({
  accessToken: "test",
  appSecret: "test",
  phoneNumberId: "test",
  verifyToken: "test",
  userName: "test",
  logger: new ConsoleLogger("silent"),
});

console.log(adapter.renderFormatted(parseMarkdown("(~80 % easy)")));
// Expected: (~80 % easy)
// Actual:   (\~80 % easy)
```

### Chat SDK Version

chat@4.37.0; @chat-adapter/whatsapp@4.39.0

### Node.js Version

Not used for reproduction — Bun 1.4.0

### Platform Adapter

WhatsApp

### Operating System

macOS

### Additional Context

The issue appears to originate in `packages/adapter-whatsapp/src/markdown.ts`:

1. `WhatsAppFormatConverter.fromAst()` calls the shared `stringifyMarkdown()` helper.
2. That helper uses `remark-stringify` with `remark-gfm`. Its strikethrough serializer escapes literal `~` characters.
3. `toWhatsAppFormat()` converts bold and strikethrough markers but leaves the Markdown escapes intact.

The whole-string regex conversions also modify formatting markers inside code: `**literal**` in a fenced code block becomes `*literal*`.

v01dst:
## Summary

Fixes #908

The WhatsApp formatter was delivering serializer-escaped tildes to users: text like `(~80 % easy)` arrived as `(\\~80 % easy)` because `remark-stringify` escapes literal `~` characters and `toWhatsAppFormat` left those escapes intact.

The whole-string marker conversions also rewrote `**bold**` and `~~strikethrough~~` inside fenced and inline code (`**literal**` in a code block became `*literal*`).

## Changes

- `toWhatsAppFormat` now processes the stringified output line by line, skipping fenced code regions, and resolves serializer-added tilde escapes only when they are not part of a real escaped backslash sequence
- Inline code spans are masked out before any conversion and restored afterwards, in both `toWhatsAppFormat` and `fromWhatsAppFormat`
- `fromWhatsAppFormat` no longer rewrites `*bold*` / `~strike~` inside inline code, so a round trip through the adapter preserves code content exactly

## Testing

- Regression tests for the reported repro: `(~80 % easy)` round trips unchanged, and multiple literal tildes stay unescaped
- Tests for intentional escapes (`\\~` stays intact), fenced code, and inline code in both directions
- Full adapter-whatsapp suite passes (235 tests) and `tsc --noEmit` is clean

cc @vercel/chat

### DIFF
--- a/packages/adapter-whatsapp/src/markdown.test.ts
+++ b/packages/adapter-whatsapp/src/markdown.test.ts
@@ -102,11 +102,43 @@ describe("WhatsAppFormatConverter", () => {
       expect(result).not.toContain("~~strikethrough~~");
     });
 
-    it("should preserve escaped asterisks and tildes as literals", () => {
+    it("should preserve escaped asterisks as literals while keeping tildes unescaped", () => {
       const ast = converter.toAst("a \\* b and c \\~ d");
       const result = converter.fromAst(ast);
       expect(result).toContain("\\*");
-      expect(result).toContain("\\~");
+      expect(result).not.toContain("\\~");
+      expect(result).toContain("~");
+    });
+
+    it("should not escape literal tildes in plain text", () => {
+      const ast = converter.toAst("(~80 % easy)");
+      const result = converter.fromAst(ast);
+      expect(result).toBe("(~80 % easy)");
+    });
+
+    it("should not escape multiple literal tildes", () => {
+      const ast = converter.toAst("run ~45 min\n\npace ~6:29/km");
+      const result = converter.fromAst(ast);
+      expect(result).toBe("run ~45 min\n\npace ~6:29/km");
+    });
+
+    it("should preserve a literal backslash before a tilde", () => {
+      const ast = converter.toAst("a \\\\~ b");
+      const result = converter.fromAst(ast);
+      expect(result).toBe("a \\\\~ b");
+    });
+
+    it("should not rewrite markers inside fenced code", () => {
+      const ast = converter.toAst("x\n\n```\n**keep** ~y~\n```\n\nz");
+      const result = converter.fromAst(ast);
+      expect(result).toContain("```\n**keep** ~y~\n```");
+      expect(result).toContain("\nz");
+    });
+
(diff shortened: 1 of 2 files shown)


v01dst:
Closing this. I was batch-generating PRs across repos while focused on my own project — that wasn't fair on reviewers' time, and I've stopped. Apologies for the noise.

PR #939, opened 2026-09-15
feat(gmail): add email adapter and standalone primitives

dancer:
## summary

- add `@chat-adapter/gmail` with standalone `/api`, `/format` and `/webhook` primitives that do not load the Chat runtime
- support label-based intake through authenticated Pub/Sub notifications, history synchronization and threaded replies in the optional root adapter
- preserve the native Pub/Sub envelope and expose all history change types, mailbox search, explicit pagination and optional whole-mailbox watches for integrations that own their sessions
- return message pointers without implicit hydration so callers choose which emails to load
- expose message reads, sending, drafts, attachments, label lookup and serializable reply context
- default to sender-only replies, with opt-in reply-all and separate direct emails through the existing private-delivery fallback
- buffer streamed replies into one complete email, without pretending sent email is editable or ephemeral
- normalize safe numeric notification history ids to strings, preserve synchronization cursors and suppress duplicate processing with persistent state
- isolate oversized, malformed and failed handoffs without automatically replaying potentially completed sends
- document mailbox authorization, cursor recovery, acknowledgement and retry ownership without adding session or worker behavior to the primitives

## verified

- full local validation, Gmail regression tests and runtime-free subpath import checks
- real Gmail delivery, attachments, reply threading, streaming and follow-ups, alongside Redis-backed restart and controlled failure checks
- native mailbox search and history changes against Google, with browser checks in both test mailboxes
- Google-authenticated Pub/Sub delivery through a real local agent runtime to a real-model reply in the correct Gmail thread
- notification replay suppression with the real model, plus session restart and follow-up reply metadata verified separately with a deterministic model


packages/adapter-gmail/src/api.ts
```diff
@@ -0,0 +1,302 @@
...
+  options: GmailApiOptions
+): Promise<GmailLabels> {
+  return request(options, "labels", labels);
+}
+
+export function stopGmailMailbox(options: GmailApiOptions): Promise<void> {
+  return request(
+    options,
+    "stop",
+    z
```
dvoytenko: put in const and similar cases too

packages/adapter-gmail/src/api.ts
```diff
@@ -0,0 +1,302 @@
...
+
+async function request<T>(
+  options: GmailApiOptions,
+  path: string,
+  result: z.ZodType<T>,
+  body?: unknown,
+  method: "GET" | "POST" = body === undefined ? "GET" : "POST"
+): Promise<T> {
+  options.signal?.throwIfAborted();
+  const account = mailbox.parse(options.mailbox);
```
dvoytenko: This is often defaultable to `"me"`, right?

packages/adapter-gmail/src/index.ts
```diff
@@ -0,0 +1,601 @@
...
+  private readonly verifier: ReturnType<typeof createGmailWebhookVerifier>;
+  private readonly logger;
+  private readonly formatter = new GmailFormatConverter();
+  private readonly context = new AsyncLocalStorage<GmailEmail>();
+  private synchronizer?: GmailSynchronizer;
+
+  constructor(config: GmailAdapterConfig = {}) {
+    try {
+      this.userName = mailbox.parse(
+        config.mailbox ?? process.env.GMAIL_MAILBOX
```
dvoytenko, line 103: Are we explicitly disallowing "me" here? If so, let's throw a validation error right away, since "me" is a very common default.

packages/adapter-gmail/src/index.ts
```diff
@@ -0,0 +1,601 @@
...
+            config.clientSecret ?? process.env.GMAIL_CLIENT_SECRET ?? "",
+          refreshToken:
+            config.refreshToken ?? process.env.GMAIL_REFRESH_TOKEN ?? "",
+          fetch: config.fetch,
+        });
+      if (typeof token === "string") {
+        z.string().min(1).parse(token);
+      }
+      this.api = { mailbox: this.userName, token, fetch: config.fetch };
+      this.verifier = createGmailWebhookVerifier({
```
dvoytenko, line 129: We also often allow a `webhookVerifier` function override in the shape of `webhookVerifier?: (req) => unknown | Promise<unknown>`. Take a look in Slack adapter.

packages/adapter-gmail/src/index.ts
```diff
@@ -0,0 +1,601 @@
...
+    if (!this.synchronizer) {
+      throw new ValidationError(
+        "gmail",
+        "Initialize Chat before using Gmail synchronization"
+      );
+    }
+    return this.synchronizer;
+  }
+
+  watch(topicName = this.topicName) {
```
dvoytenko, line 182: Do we re-request from the saved point right before or right after calling `watch`? Or expecting a manual call to `sync`?

packages/adapter-gmail/src/index.ts
```diff
@@ -0,0 +1,601 @@
...
+    if (!topicName) {
+      throw new ValidationError(
+        "gmail",
+        "Provide topicName or GMAIL_TOPIC_NAME to register a mailbox watch"
+      );
+    }
+    return this.call(() => this.synchronization.watch(topicName));
```
dvoytenko, line 189: Do we need to worry about Google automatically stopping the watch? I think it does a 24hour subscriptions only?

packages/adapter-gmail/src/index.ts
```diff
@@ -0,0 +1,601 @@
...
+    return decodeGmailThread(value, this.userName);
+  }
+
+  channelIdFromThreadId(value: string): string {
+    this.decodeThreadId(value);
+    return gmailChannel(this.userName);
+  }
+
+  getChannelVisibility() {
+    return "private" as const;
```
dvoytenko: why?

packages/adapter-gmail/src/index.ts
```diff
@@ -0,0 +1,601 @@
...
+      isDM: false,
+      metadata: { mailbox: this.userName, threadId: result.id },
+    };
+  }
+
+  renderFormatted(content: FormattedContent): string {
+    return this.formatter.fromAst(content);
+  }
+
+  async postMessage(
```
dvoytenko, line 457: FMI: does this Reply All? 

I believe we also need `postEmphemeral` support, especially with `fallbackToDM: true` for authorizations and similar private-only events. I think alternative would be to also support the sequence of openDM() + postMessage() for these.

packages/adapter-gmail/src/sync.ts
```diff
@@ -0,0 +1,322 @@
...
+            }
+            throw error;
+          }
+        );
+        if (!thread) {
+          continue;
+        }
+        ordered = thread.messages
+          .map((message) => message.id)
+          .filter((id) => selected.has(id));
```
dvoytenko, line 231: Why do this and not just load messages based on the references? This is very likely over-fetching significantly - basically the whole thread for every new message. Is this for ordering only? Doesn't gmail give us that via history order?

packages/adapter-gmail/README.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+The root adapter currently uses label-based handoff. It coalesces selected emails from the same conversation within one synchronization run into the newest incoming message. Older messages remain available through thread history. Use durable state in production; in-memory state is only suitable for local experiments.
+
+The selected label controls dispatch, not OAuth access to the mailbox. `thread.post()` sends email immediately. Streaming buffers text and sends once; sent email cannot be edited or retracted. Bridges that implement their own post/edit streaming must disable that behavior for email.
+
+## setup
+
+1. Enable the Gmail API and Pub/Sub in your Google Cloud project. Obtain user-context OAuth credentials for the mailbox. Request `gmail.readonly` for reading and watches, plus `gmail.send` for replies. Draft creation needs `gmail.compose`. The adapter does not need permission to delete email or modify labels. See [Gmail scopes](https://developers.google.com/workspace/gmail/api/auth/scopes).
+2. Create a Pub/Sub topic in the same Google Cloud project as the OAuth client. Grant `gmail-api-push@system.gserviceaccount.com` permission to publish to that topic.
```
dvoytenko, line 159: might be worth a couple quick commands, such as

```
gcloud pubsub topics create gmail-events
gcloud pubsub topics add-iam-policy-binding gmail-events \
  --member="serviceAccount:gmail-api-push@system.gserviceaccount.com" \
  --role="roles/pubsub.publisher"
gcloud pubsub subscriptions create gmail-webhook \
  --topic=gmail-events \
  --push-endpoint=...
```

packages/adapter-gmail/README.md
```diff
@@ -0,0 +1,158 @@
...
+| `pubsubAudience` | `GMAIL_PUBSUB_AUDIENCE` |
+| `pubsubServiceAccountEmail` | `GMAIL_PUBSUB_SERVICE_ACCOUNT_EMAIL` |
+| `subscription` | `GMAIL_SUBSCRIPTION` |
+| `topicName` | `GMAIL_TOPIC_NAME` |
+
+Use either the OAuth client/refresh-token configuration or `accessToken`. A token resolver is recommended when another system already manages refresh. Explicit OAuth configuration takes precedence over an environment access token. The mailbox is an email address, the subscription is `projects/PROJECT/subscriptions/SUBSCRIPTION`, and the topic is `projects/PROJECT/topics/TOPIC`.
+
+```typescript
+await bot.initialize();
+await gmail.watch();
```
dvoytenko, line 212: `await gmail.sync()` right away? AFAICT `watch` doesn't "catch up" to recent history.

dvoytenko (changes requested):

packages/adapter-gmail/src/sync.ts
dancer: this is only for ordering multiple selected messages in one thread. it fetches IDs only, not bodies, and skips the lookup for a single message. history orders mailbox changes, so labelling an older email can put it last

dvoytenko (approved):

bensabic (approved):

packages/adapter-gmail/src/index.ts
```diff
@@ -0,0 +1,699 @@
...
+    }
+    const recipient = this.decodeThreadId(await this.openDM(userId));
+    const thread =
+      threadId === gmailChannel(this.userName)
+        ? recipient
+        : this.decodeThreadId(threadId);
+    const continuation =
+      thread.threadId === undefined
+        ? undefined
+        : await this.continuation(thread.threadId);
```
dvoytenko, line 443: Do we need to pass `replayAll: false` override here somehow?

dvoytenko (changes requested):

dvoytenko (approved):

